PR #3, opened 2020-07-24
Add API sketch and clarify data structures.

danyao:
rouslan@ PTAL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,4 +179,4 @@ If the payment instrument specified by `instrumentId` is not available or if the
 user agent will open `fallbackUrl` inside the Secure Modal Window.
 
 [PublicKeyCredential]: https://www.w3.org/TR/webauthn/#iface-pkcredential
-[WebAuthentication]: https://www.w3.org/TR/webauthn/
+[Web Authentication]: https://www.w3.org/TR/webauthn/
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ dictionary SecurePaymentCredentialCreationOptions {
 };
 
 dictionary SecurePaymentCredentialInstrument {
+  // |instrumentId| is a caller provided ID for the payment instrument to which the new PaymentCredential should
+  // be bound. It should be an opaque string generated using a payment network specific algorithm that allows the
+  // network to identify the issuer of the instrument and the issuer to identify the account associated with this
+  // instrument.
   required DOMString instrumentId;
   required DOMString displayName;
   required USVString icon;
@@ -50,7 +54,7 @@ Example usage:
 const securePaymentConfirmationCredentialCreationOptions = {
   instrument: {
     instrumentId: "Q1J4AwSWD4Dx6q1DTo0MB21XDAV76",
-    displayName: 'Mastercard****4444',
+    displayName: 'Mastercard····4444',
     icon: 'icon.png'
   },
   challenge,
@@ -87,7 +91,7 @@ Example usage:
 const securePaymentConfirmationCredentialCreationOptions = {
   instrument: {
     instrumentId: "Q1J4AwSWD4Dx6q1DTo0MB21XDAV76",
-    displayName: 'Mastercard****4444',
+    displayName: 'Mastercard····4444',
     icon: 'icon.png'
   },
   existingCredential: {
@@ -131,7 +135,7 @@ const credential = await navigator.credentials.get({
 
 ## Authenticating a payment
 
-Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a SecurePaymentCredential created by any other origin. The `PaymentRequest.show()` method must require a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
+Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a SecurePaymentCredential created by any other origin. The `PaymentRequest.show()` method requires a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
 
 Proposed new `secure-payment-confirmation` payment method:
 
@@ -139,6 +143,8 @@ Proposed new `secure-payment-confirmation` payment method:
 dictionary SecurePaymentConfirmationRequest {
   SecurePaymentConfirmationAction action;
   required DOMString instrumentId;
+  // Opaque data about the current transaction provided by the issuer. As the issuer is the RP
+  // of the credential, |networkData| provides protection against replay attacks.
   required BufferSource networkData;
   unsigned long timeout;
   required USVString fallbackUrl;
@@ -149,9 +155,15 @@ enum SecurePaymentConfirmationAction {
 };
 
 dictionary SecurePaymentConfirmationResponse {
+  // A bundle of key attributes about the transaction.
+  required SecurePaymentConfirmationPaymentData paymentData;
+  // The WebAuthn assertion created by signing over |paymentData|.
+  required AuthenticatorAssertionResponse assertion;
+};
+
+dictionary SecurePaymentConfirmationPaymentData {
   required SecurePaymentConfirmationMerchantData merchantData;
   required BufferSource networkData;
-  required AuthenticatorAssertionResponse assertion;
 };
 
 dictionary SecurePaymentConfirmationMerchantData {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,30 +129,60 @@ const credential = await navigator.credentials.get({
...
-Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) to prompt the user to verify a credential created by any other origin. The `PaymentRequest.show()` method must require a user gesture and display user interface with the amount of payment and the hostname of the top-level context where the `PaymentRequest` API was invoked.
+Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a SecurePaymentCredential created by any other origin. The `PaymentRequest.show()` method must require a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
+
+Proposed new `secure-payment-confirmation` payment method:
+
+```webidl
+dictionary SecurePaymentConfirmationRequest {
+  SecurePaymentConfirmationAction action;
+  required DOMString instrumentId;
+  required BufferSource networkData;
```
danyao, line 142: @btidor-stripe: Do you think the `networkData` needs to be specified as part of Secure Payment Confirmation or can we leave it opaque, and let the underlying payment rails (e.g. 3DS in the current use case) dictate what goes into it?

Since the browser will just be hashing this data and then getting the authenticator to sign over it, it keeps things simple if the detailed structure is left out of the spec. It also makes it more extensible to other payment rails in the future.

README.md
```diff
@@ -54,30 +129,60 @@ const credential = await navigator.credentials.get({
...
+  unsigned long timeout;
+  required USVString fallbackUrl;
+};
+
+enum SecurePaymentConfirmationAction {
+  "authenticate",
```
rsolomakhin, line 148: Do we expect more actions in the future?

README.md
```diff
@@ -54,30 +129,60 @@ const credential = await navigator.credentials.get({
...
   }],
   {total: {label: 'total', amount: {currency: 'USD', value: '20.00'}}});
 const response = await request.show();
+
+// Merchant validates |response.merchantData| and sends |response| to issuer for authentication.
 ```
 
-If none of the credentials from `allowCredentials.id` are available, then the user agent will invoke the payment handler for the payment method specified in the `fallback` method. It is recommended that the `fallback` payment method supports just-in-time installation of a payment handler, so the user agent may install it just-in-time. This enables the payment handler to open a fallback URL in the payment handler window.
+If the payment instrument specified by `instrumentId` is not available or if the user failed to authenticate, then the
+user agent will open `fallbackUrl` inside the Secure Modal Window.
```
rsolomakhin, line 185: Secure Modal Window is a UI paradigm that does not define any communication channel to the merchant or the browser at this time. Should `PaymentRequest.show()` return some kind of indicator to the merchant that this has happened and the browser will not be showing an authentication prompt to the user?

README.md
```diff
@@ -4,45 +4,120 @@ The goal is to improve payment confirmation experience with the help of WebAuthn
...
+
 ## Creating a credential
 
-A new credential type is introduced for `navigator.credentials.create()` that stores the name and the icons of a payment instrument.
+A new `SecurePaymentCredential` credential type is introduced for `navigator.credentials.create()` to bind descriptions of a payment instrument, i.e. a name and an icon, with a vanilla [PublicKeyCredential].
+
+Proposed new spec that extends [Web Authentication]:
+```webidl
+[SecureContext, Exposed=Window]
+interface SecurePaymentCredential : PublicKeyCredential {
```
marcoscaceres, line 20: Could we drop the "Secure" bit? Seems redundant.

```suggestion
interface SecurePaymentCredential : PublicKeyCredential {
```

README.md
```diff
@@ -4,45 +4,120 @@ The goal is to improve payment confirmation experience with the help of WebAuthn
...
-    name: 'Mastercard****4444',
-    icons: [{
-  	  'src': 'icon.png',
-  	  'sizes': '48x48',
-  	  'type': 'image/png',
-    }],
+const securePaymentConfirmationCredentialCreationOptions = {
+  instrument: {
+    instrumentId: "Q1J4AwSWD4Dx6q1DTo0MB21XDAV76",
+    displayName: 'Mastercard****4444',
```
marcoscaceres, line 53: I'm worried about these kinds of display names... they are somewhat easy to screw up (and using "****" is a bit ugly). Figuring out how to obfuscate the display of a payment credential is something we need to figure out.

README.md
```diff
@@ -4,45 +4,120 @@ The goal is to improve payment confirmation experience with the help of WebAuthn
...
+  // user: For a SecurePaymentCredential, |instrument| is analogous to |user|.
+  // excludeCredentials: No payment use case has been proposed for this field.
+  // attestation: Authenticator attestation is considered an anti-pattern for adoption so will not be supported.
+  // extensions: No payment use case has been proposed for this field.
+};
+
+dictionary SecurePaymentCredentialInstrument {
+  required DOMString instrumentId;
+  required DOMString displayName;
+  required USVString icon;
```
marcoscaceres, line 44: one icon might not be sufficient... the icon might get displayed in multiple contexts - like in a payment sheet or in a keyboard. In a keyboard, it would be maybe 32x32 pixels, so will need special treatment (or an entirely new icon) to show properly when compared to the one shown in a sheet (or even in a minimal UI, which may be say 100x100px)

README.md
marcoscaceres: the "displayName" might need to be structured somehow... we might need to look at all sorts of instruments to figure out common bits of information.

rsolomakhin: Unicode has support for better looking characters, such as the [middle dot](https://en.wikipedia.org/wiki/Interpunct), which is used in Chrome UI to obfuscate credit cards, the format of which is well known. Callers of this API can use the middle dot as well, especially if we put it in public documentation.

My hope is this API will be used for more than only credit cards, so the best obfuscation is hard to guess ahead of time.

README.md
```diff
@@ -54,30 +129,60 @@ const credential = await navigator.credentials.get({
 });
 ```
 
-## Confirming payment
+## Authenticating a payment
 
-Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) to prompt the user to verify a credential created by any other origin. The `PaymentRequest.show()` method must require a user gesture and display user interface with the amount of payment and the hostname of the top-level context where the `PaymentRequest` API was invoked.
+Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a SecurePaymentCredential created by any other origin. The `PaymentRequest.show()` method must require a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
```
marcoscaceres, line 134: > The `PaymentRequest.show()` method must require a user gesture.

That's a given, as it's baked into Payment Request... maybe just a statement of fact: "The `PaymentRequest.show()` method requires a user gesture."

README.md
danyao: Sure!

README.md
danyao: > My hope is this API will be used for more than only credit cards, so the best obfuscation is hard to guess ahead of time.

+1. The caller of this API is an issuer and they are incentivized to figure out the best way to obfuscate the payment instrument ID and with unicode, has the tools to generate good looking strings. IMHO, imposing more structure here has the cost of less flexibility and tighter coupling between browser and the payment instrument provider, with very small UX gain.

README.md
danyao: Yes. "select-and-authenticate" may be a future action for a combined instrument selection and authenticate flow. See https://docs.google.com/document/d/1n73fdsn8tTDYUGjQPOGRGI_VcTaGNjweedSgcH7P8Hc/edit?ts=5ef93da4&pli=1#heading=h.y7uxuzpcwy0b

README.md
rsolomakhin: We'd like to run this as an origin trial on short timeframe, so let's go with something that's easiest to implement: a single URL. Origin trials have explicit agreement that the user agent is free to break the API after the origin trial expires, so we can enable usage of multiple icons in the future. Also, I believe there's some precedent in WebAuthn for using a single image URL.

README.md
danyao: There's an open question for a payment handler should be involved in a "secure-payment-confirmation" transaction. If it is, then the most natural way is probably to expose the error via an extension to `PaymentRequestEvent`. But if a payment handler should not be involved, maybe the browser should build a new communication channel for the web content inside a Secure Modal Window to communicate with the browser. More discussion around this is needed.

For now, I just called out the open question, and left a suggested hack for the pilot.

README.md
btidor-stripe: Agreed! I don't think the browser needs to know anything about what goes into `networkData` -- my thought was that payment systems might suggest some or all of its contents, the same way they currently define standard cryptogram formats.

PR #9, opened 2020-07-29
Payment Handlers

adrianhopebailie:
Proposal for using SCP with Payment Handlers

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The rest of the document is organized into these sections:
 
 - [Creating a credential](#creating-a-credential)
 - [Querying a credential](#querying-a-credential)
-- [Authenticating a payment](#authentication-a-payment)
+- [Confirming a payment](#confirming-a-payment)
 
 ## Creating a credential
 
@@ -133,9 +133,25 @@ const credential = await navigator.credentials.get({
 });
 ```
 
-## Authenticating a payment
+## Confirming a payment
 
-Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a `PaymentCredential` created by any other origin. The `PaymentRequest.show()` method requires a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
+Any origin may invoke the [Payment Request API] in order to use a PaymentCredential to confirm a payment. The caller will either:
+ - invoke the API using a known `instrumentId` (using the `secure-payment-confirmation` payment method identifier), or
+ - invoke the API using one or more payment methods that support a Secure Payment Confirmation response and have a supporting Payment Handler installed
+
+**Note:** The `PaymentRequest.show()` method requires a user gesture.
+
+### Using the `secure-payment-confirmation` payment method
+
+Any origin can use the [Payment Request API] with the `secure-payment-confirmation` payment method to immediately prompt the user to use a `PaymentCredential` to confirm a payment.
+
+In this case the caller must provide:
+ - an `instrumentId` provided by the RP for a payment instrument that was already selected by the user
+ - an opaque `challenge` provided by the RP
+ 
+These data elements will be sourced directly via the payment network by the caller before calling the API. This flow is well suited to payment methods such as card payments with 3D Secure (3DS) where the user has entered a payment card number into a form or selected a card on file (i.e. selected a payment instrument) and the caller has used the 3DS infrastructure to get a valid `instrumentId` and dynamic `networkData` from the card issuer.
+ 
+After calling `PaymentRequest.show()` the browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
 
 Proposed new `secure-payment-confirmation` payment method:
 
@@ -144,8 +160,8 @@ dictionary SecurePaymentConfirmationRequest {
   SecurePaymentConfirmationAction action;
   required DOMString instrumentId;
   // Opaque data about the current transaction provided by the issuer. As the issuer is the RP
-  // of the credential, |networkData| provides protection against replay attacks.
-  required BufferSource networkData;
+  // of the credential, |challenge| provides protection against replay attacks.
+  required BufferSource challenge;
   unsigned long timeout;
   required USVString fallbackUrl;
 };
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -186,9 +186,25 @@ const credential = await navigator.credentials.get({
...
+
+### Using the `secure-payment-confirmation` payment method
+
+Any origin can use the [Payment Request API] with the `secure-payment-confirmation` payment method to immediately prompt the user to use a `PaymentCredential` to confirm a payment.
+
+In this case the caller must provide:
+ - an `instrumentId` provided by the RP for a payment instrument that was already selected by the user
+ - an opaque `challenge` provided by the RP
+ 
+These data elements will be sourced directly via the payment network by the caller before calling the API. This flow is well suited to payment methods such as card payments with 3D Secure (3DS) where the user has entered a payment card number into a form or selected a card on file (i.e. selected a payment instrument) and the caller has used the 3DS infrastructure to get a valid `instrumentId` and dynamic `networkData` from the card issuer.
```
danyao, line 205:
```suggestion
These data elements will be sourced directly via the payment network by the caller before calling the API. This flow is well suited to payment methods such as card payments with 3D Secure (3DS) where the user has entered a payment card number into a form or selected a card on file (i.e. selected a payment instrument) and the caller has used the 3DS infrastructure to get a valid `instrumentId` and dynamic `challenge` from the card issuer.
```

README.md
```diff
@@ -186,9 +186,25 @@ const credential = await navigator.credentials.get({
 });
 ```
 
-### Authenticating a payment
+## Confirming a payment
 
-Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a `PaymentCredential` created by any other origin. The `PaymentRequest.show()` method requires a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
+Any origin may invoke the [Payment Request API] in order to use a PaymentCredential to confirm a payment. The caller will either:
```
danyao, line 191:
```suggestion
Any origin may invoke the [Payment Request API] in order to use a `PaymentCredential` to confirm a payment. The caller will either:
```

README.md
```diff
@@ -252,12 +268,58 @@ If the payment instrument specified by `instrumentId` is not available or if the
...
+ 3. If the set of user choices is not empty the user is prompted to select a payment instrument, else the browser reverts to the usual Payment Handler selection flow.
+ 4. If the user selects an instrument the corresponding Payment Handler is invoked and the `instrumentId` is passed to it as the [`instrumentKey`](https://w3c.github.io/payment-handler/#dom-paymentrequestevent-instrumentkey) property of the `PaymentRequestEvent`.
+ 
+ If the user's chosen instrument can be used for Secure Payment Confirmation (i.e. is linked to a credential) `PaymentRequestEvent.isSecureConfirmationAvailable` will be `true`.
+ 
+ The Payment Handler can can query its backend systems to get the neccessary challenge (`challenge`) and verify that the `instrumentId` is still valid before calling `PaymentRequestEvent.securePaymentConfirmation(challenge, timeout)` to invoke the Secure Payment Confirmation flow.
+ 
+The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
```
danyao, line 286: Can we move this step before triggering the `PaymentRequestEvent`? I think functionally the flow will be identical.

To mitigate the privacy threat of sharing data to a 3rd party before user awareness, I'd like to make sure all new flows we design are compatible with the "native UI first" principle. Browser will only release data to the payment handler if the user is aware that they're about to start a payment by clicking "Verify" on the native browser UI.

README.md
```diff
@@ -252,12 +268,58 @@ If the payment instrument specified by `instrumentId` is not available or if the
...
+  readonly attribute object? paymentOptions;
+  readonly attribute FrozenArray<PaymentShippingOption>? shippingOptions;
+  Promise<WindowClient?> openWindow(USVString url);
+  Promise<PaymentRequestDetailsUpdate?> changePaymentMethod(DOMString methodName, optional object? methodDetails = null);
+  Promise<PaymentRequestDetailsUpdate?> changeShippingAddress(optional AddressInit shippingAddress = {});
+  Promise<PaymentRequestDetailsUpdate?> changeShippingOption(DOMString shippingOption);
+  void respondWith(Promise<PaymentHandlerResponse> handlerResponsePromise);
+  // New members
+  readonly attribute boolean isSecurePaymentConfirmationAvailable;
+  Promise<SecurePaymentConfirmationResponse?> securePaymentConfirmation(BufferSource challenge, long timeout);
```
danyao, line 311:
```suggestion
partial interface PaymentRequestEvent {
  readonly attribute boolean isSecurePaymentConfirmationAvailable;
  Promise<SecurePaymentConfirmationResponse?> securePaymentConfirmation(BufferSource challenge, long timeout);
```

README.md
```diff
@@ -186,9 +186,25 @@ const credential = await navigator.credentials.get({
...
+
+**Note:** The `PaymentRequest.show()` method requires a user gesture.
+
+### Using the `secure-payment-confirmation` payment method
+
+Any origin can use the [Payment Request API] with the `secure-payment-confirmation` payment method to immediately prompt the user to use a `PaymentCredential` to confirm a payment.
+
+In this case the caller must provide:
+ - an `instrumentId` provided by the RP for a payment instrument that was already selected by the user
+ - an opaque `challenge` provided by the RP
```
btidor-stripe, line 203: Hi @adrianhopebailie, this shouldn't be necessary: we're hashing the `paymentData` to get the challenge and there's nothing in there that has to be provided by the RP.

adrianhopebailie: We can't invoke WebAuthn without some dynamic data from the RP.

README.md
adrianhopebailie: A few issues/comments:

 1. It's going to be a problem in future if the UI is combined and the payment details are rendered by the authenticator as we'll already need the challenge from the RP.
 2. Assuming the user has already selected an instrument they have already been presented with a native UI and implicitly chosen to interact with the issuer of that instrument.
 3. If the instrument data is stale the user may have chosen an instrument that subsequently gets rejected by the RP when the PH is invoked and goes online to get the challenge.

adrianhopebailie:
Closing as this is now stale. Working on a more comprehensive approach [here](https://github.com/w3c/webpayments/blob/gh-pages/proposals/arch2020.md).

PR #36, opened 2020-11-02
FAQ: Can the party that calls SPC validate the assertion even if not the RP?

ianbjacobs:
Based on conversations about SPC I thought it might be useful to start an FAQ.

I started it with a first question: 
 Can the party that calls SPC validate the assertion even if not the RP?

Ian

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,9 +290,9 @@ The authenticator signature is emphemeral to each transaction and is not useful 
 
 ## FAQ
 
-### Q. Can the party that calls SPC validate the assertion even if not the RP?
+### Q. Who can validate the SPC response besides the actual Relying Party (RP)?
 
-Yes. The Relying Party can share the credentialPublicKey (of the authenticator data) with another party, e.g., through an out-of-band communication.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -288,6 +288,11 @@ The authenticator signature is emphemeral to each transaction and is not useful
 [Web Authentication]: https://www.w3.org/TR/webauthn/
 [pilot]: https://bit.ly/webauthn-to-pay-2020h2-pilot
 
+## FAQ
+
+### Q. Can the party that calls SPC validate the assertion even if not the RP?
```
danyao, line 293:
```suggestion
### Q. Who can validate the SPC assertion besides the actual Relying Party (RP)?
```

README.md
```diff
@@ -288,6 +288,11 @@ The authenticator signature is emphemeral to each transaction and is not useful
...
+
+Yes. The Relying Party can share the credentialPublicKey (of the authenticator data) with another party, e.g., through an out-of-band communication.
```
danyao, line 295:
```suggestion
Validation in the context of SPC means the ability to verify that the signature in the SPC response is a signature over the SPC challenge (which bundles together transaction details with transaction-specific dynamic data from the RP). An RP can choose to share the credential public key with another party (e.g. the card network or PSP) via out-of-band communication to enable said party to validate the SPC assertion.
```

danyao (commented):
@ianbjacobs Tweaked the wording a bit. WDYT?

ianbjacobs:
@danyao,
Thanks for the suggestions. I took your text, moved the bits around, and updated the proposal.

PR #64, opened 2021-05-04
Update explainer with cross-origin notes

stephenmcgruer:
Both for enrollment and authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,8 @@ cross-origin iframe (e.g. if `merchant.com` embeds an iframe from `bank.com`). T
 intended to support the common enrollment flow of a bank enrolling the user during a
 step-up challenge (e.g. after proving their identity via OTP).
 
-To allow this, the cross-origin iframe must have `allow="payment"` set:
+To allow this, the cross-origin iframe must have the ["payment" permission policy](
+https://w3c.github.io/payment-request/#permissions-policy) set. For example:
 
 ```html
 <!-- Assume parent origin is merchant.com -->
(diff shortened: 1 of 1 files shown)


ianbjacobs (approved):

README.md
```diff
@@ -121,6 +121,21 @@ const credential = await navigator.credentials.create({
...
 See the [Guide to Web Authentication](https://webauthn.guide/) for mode details about the `navigator.credentials` API.
 
+#### Creating a PaymentCredential in a cross-origin iframe
+
+Unlike normal WebAuthn credentials, a PaymentCredential is allowed to be created in a
+cross-origin iframe (e.g. if `merchant.com` embeds an iframe from `bank.com`). This is
+intended to support the common enrollment flow of a bank enrolling the user during a
+step-up challenge (e.g. after proving their identity via OTP).
+
+To allow this, the cross-origin iframe must have `allow="payment"` set:
```
rsolomakhin, line 131: Nit: Technically speaking, `allow="payment"` is only one way to set permission policy, so stating "must have `allow="payment"`" can be misleading. Why not state that the "iframe must have the "payment" permission policy set, for example:"... and then follow up with the `<iframe src="https://example.org" allow="payment">` example as you have below. Do you think that would be more accurate?

README.md
```diff
@@ -186,6 +201,11 @@ const credential = await navigator.credentials.get({
 
 Any origin may invoke the [Payment Request API](https://w3c.github.io/payment-request/) with the `secure-payment-confirmation` payment method to prompt the user to verify a `PaymentCredential` created by any other origin. The `PaymentRequest.show()` method requires a user gesture. The browser will display a native user interface with the payment amount and the payee origin, which is taken to be the origin of the top-level context where the `PaymentRequest` API was invoked.
 
+Note that [per the Payment Request specification](https://www.w3.org/TR/payment-request/#using-with-cross-origin-iframes),
+if `PaymentRequest` is used within a cross-origin iframe (e.g. if `merchant.com`
+embeds an iframe from `psp.com`, and `psp.com` wishes to use `PaymentRequest`),
+that iframe must have an `allow` attribute with the `"payment"` keyword.
```
rsolomakhin, line 207: Same comment as above.

rsolomakhin (changes requested):
Should we clarify that allow="payment" is not the only way to set the permission policy? For example, permission policy can be set in HTTP headers. Another example is specifying the origins, that is `allow="payment https://example.com"`. For more details, see:

https://github.com/w3c/webappsec-permissions-policy/blob/main/permissions-policy-explainer.md#how-is-a-policy-specified

README.md
stephenmcgruer: Not a nit at all; a learning lesson for me (I didn't know about the other ways to set permission policy). Done.

rsolomakhin (approved):

PR #70, opened 2021-05-11
Clearer benefit/feature descriptions

ianbjacobs:
Per 10 May SPC task force discussion, this PR splits "benefits" from "features".

### DIFF
--- a/scope.md
+++ b/scope.md
@@ -25,32 +25,29 @@ SPC is a tool for this last step.
 
 ### Benefits
 
-* **Faster Authentication**. FIDO authentication provides a very good
-user experience. Our expectation is that the SPC "happy path" further
-accelerates strong customer authentication compared to other
-approaches such as one-time passwords. An important goal is to remove
-the need to redirect an enrolled user during a payment
-transaction. Another important goal is to speed up authentication
-compared to other approaches such as one-time passwords. A third goal
-is to create a consistent and trusted authentication experience across
-the Web.
+* **Authentication Streamlined for Payment**. FIDO authentication
+provides a very good user experience. We expect SPC to build on that
+experience in several ways, by further accelerating authentication
+(compared to one-time passcodes), requiring fewer user gestures,
+offering a predictable user experience across sites, and avoiding
+redirects.
 
 * **Scalable and Ubiquitous**. SPC supports streamlined authentication
 across merchant sites without additional enrollment.
 
-* **Designed to Meet Regulatory Requirements**. SPC helps entities
+* **Designed to Meet Regulatory Requirements**. The standardized
+payment confirmation user experience is designed to help entities
 fulfill regulatory requirements (e.g., strong customer authentication
-and transaction confirmation under PSD2) and other customer
+and dynamic linking under PSD2) and other customer
 authentication use cases.
 
 * **Simpler and more Secure Front-end Deployment**. The browser (or
-secure hardware) manages the display of transaction confirmation,
+secure hardware) manages the display of the payment confirmation experience,
 removing the need for other parties (e.g., issuing banks or payment
 apps) to do so. In addition, enabling payment service providers or
 others to authenticate the user can reduce the need to embed code
 provided by a Relying Party in a Web page, reducing security risks.
-Because the browser (or secure hardware) controls the display of
-transaction details, SPC is phishing-resistant.
+Reducing the need for redirects should also simplify solutions.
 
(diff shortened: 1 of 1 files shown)


Goosth:
Hi @ianbjacobs,

Sorry for the slow response here. Looking to get to this tomorrow.

scope.md
```diff
@@ -20,46 +20,58 @@ capabilities could help streamline the user experience:
...
-even more streamlined experience. One important goal is to remove the
-need to redirect an enrolled user during a payment
+### Benefits
+
+* **Faster Authentication**. FIDO authentication provides a very good
+user experience. Our expectation is that the SPC "happy path" further
+accelerates strong customer authentication compared to other
+approaches such as one-time passwords. An important goal is to remove
+the need to redirect an enrolled user during a payment
 transaction. Another important goal is to speed up authentication
```
Goosth, line 33: The description text seems to focus more the acceleration and enablement of SCA and preventing a controlled experience. For me the title (Faster Authentication) is inaccurate. Fido is already just one click. 
The challenge with it is that there is no secure display and no payment specific fields. The industry's ideal (Frictionless using Risk) has no challenge. 
What it does do is:
* Avoid the party affecting the payment to have to redirect to a third party to authenticate. So it may save a redirect. 
* Make it closer to regulatory compliance (easier to comply) since it includes a display and signature.

What if we rather make this "Controlled Authentication" or something, which would fit the description better.

scope.md
```diff
@@ -20,46 +20,58 @@ capabilities could help streamline the user experience:
...
-* **Simpler Front-end Deployment**. The browser (or secure hardware)
-manages the display of transaction confirmation, removing the need for
-other parties (e.g., issuing banks or payment apps) to do so.
+across merchant sites without additional enrollment.
+
+* **Designed to Meet Regulatory Requirements**. SPC helps entities
+fulfill regulatory requirements (e.g., strong customer authentication
+and transaction confirmation under PSD2) and other customer
+authentication use cases.
+
```
Goosth, line 45: I like this. It does seem to overlap with previous item somewhat. So suggest we remove the SCA references there.

scope.md
```diff
@@ -20,46 +20,58 @@ capabilities could help streamline the user experience:
...
+* **Simpler and more Secure Front-end Deployment**. The browser (or
+secure hardware) manages the display of transaction confirmation,
+removing the need for other parties (e.g., issuing banks or payment
+apps) to do so. In addition, enabling payment service providers or
+others to authenticate the user can reduce the need to embed code
+provided by a Relying Party in a Web page, reducing security risks.
+Because the browser (or secure hardware) controls the display of
+transaction details, SPC is phishing-resistant.
```
Goosth, line 54: Like this description. We could put the 'no need for redirect' benefit into this as well.

Unsure if SPC is phishing resistant. If the keys are retrieved by any third party (hacker-merchant.com), they would also be able to initiate this transaction from their domain.

scope.md
```diff
@@ -20,46 +20,58 @@ capabilities could help streamline the user experience:
...
 
+### Unique Features
+
+The above benefits are grounded in a small number of unique API features:
+
+* **Browser-native UX for transaction confirmation**. The browser
+provides a consistent and efficient authentication UX across merchant
+sites. Note: The architecture should support display of transaction
+confirmation details by authenticators.
+
```
Goosth, line 65: Good notes.

Small comment: We should be careful of mixing transaction and payment. SPC is Payment specific. In fact, right now  Merchant Payment specific. Transactions are typically more generic (Issue a Loan, Make a withdrawal, Deposit funds into RA, Request a Credit Increase, etc.)

scope.md
```diff
@@ -20,46 +20,58 @@ capabilities could help streamline the user experience:
...
+* **Cryptographic evidence**. Transaction confirmation generates
+cryptographic evidence of the user's confirmation of transaction
+details.
+
```
Goosth, line 69: transaction vs payment

scope.md
ianbjacobs: I've renamed the "faster" section to "Authentication Streamlined for Payment".
I've removed SCA from that one since covered under "regulatory"

scope.md
ianbjacobs: I've changed many instance of "transaction confirmation" to "payment confirmation. However, the word "transaction" still appears in the document to refer to transactions. :)

Goosth (approved):

PR #71, opened 2021-05-14
Update requirements.md

tblachowicz:
Some new requirements, some questions to the existing ones, and few corrections.

### DIFF
--- a/requirements.md
+++ b/requirements.md
@@ -16,14 +16,28 @@ See also: [SPC Scope](scope.md) for definitions and more information.
 ## Web Context Support
 
 * It must possible to invoke SPC from an iframe. See [issue 68](https://github.com/w3c/secure-payment-confirmation/issues/68).
+
+Q: I understand that woudl also require PR API to be used from within the IFrame. Is that right?
+
 * It must possible to invoke SPC from within Payment Request API.
+
+Q: Shouldn't that be rather "It must be possible to invoke SPC via Payment Request API"? I understand the intention is to use the PR API as a way to invoke the SCP.
+
 * It must possible to invoke SPC from within a Payment Handler.
+
+Q: Does is mean the Payment Handler API woudl need to be extended by some new capabilities to trigger the SPC from within the PH code?
 
 ## Enrollment
 
 * It must be possible to enroll an SPC Credential outside of a transaction.
 * It must be possible to enroll an SPC Credential during a transaction. This enrollment should not prevent timely completion of the transaction.
+
+Q: Does that also imply that the enrolment during the transaction coudl somehow produce the SPC Attestation? Such capability woudl help to solve one of the difficult problems of getting conusmer authenticated for the first transaction.
+
 * It must be possible to enroll an SPC Credential from code on a merchant site.
+
+Q: Can you elaborate more what's the use-case? My assumption was that the RP i.e. issuer or wallet would enroll the SPC Credential. The merchant or PSP woudl be leveraging the enrolled credential by triggering the SPC auth flow.
+
 * It must be possible to enroll an SPC Credential from a payment handler.
 * Each browser should natively support an SPC credential enrollment user experience.
 
@@ -31,14 +45,17 @@ See also: [SPC Scope](scope.md) for definitions and more information.
 
 * Enrollment must include display information for the associated instrument.
 * Because instrument display information is available to the relying party (e.g., provided by the relying party itself, the merchant, or some other party), it is not a requirement that this information be stored in the browser as part of the SPC Credential.
+* The relying party can update the instrument information for already enrolled SPC Credential to support the update of instrument meta-data e.g. card art.
 
 ## Transaction Confirmation User Experience
 
-* Each browser must natively support a transaction confirmation user experience.
-* Although we anticipate that in most cases the browser will render the transaction confirmation user experience, the protocol must support rendering by other entities (e.g., the operating system or authenticator).
+* Each browser must natively support a payment confirmation user experience.
+* Although we anticipate that in most cases the browser will render the payment confirmation user experience, the protocol must support rendering by other entities (e.g., the operating system or authenticator).
 * See [issue 48](https://github.com/w3c/secure-payment-confirmation/issues/48) on merchant information display.
 * For regulatory reasons, the party that invokes SPC must be able to specify a timeout for the user experience. See [issue 67](https://github.com/w3c/secure-payment-confirmation/issues/67).
-* The transaction confirmation user experience should include the beneficiary name, and optionally the title and favicon of the page where it was called. See [issue 48](https://github.com/w3c/secure-payment-confirmation/issues/48).
+* The payment confirmation user experience should include the beneficiary name, and optionally the title and favicon of the page where it was called. See [issue 48](https://github.com/w3c/secure-payment-confirmation/issues/48).
+* The payment confirmation user experience must display instrument information such as label and icon.
+* The payment confirmation user experience must display amount and currency of the payment.
 
 ### Sources of Instrument Information
 
@@ -73,8 +90,13 @@ authorization it can be validated against relying-party stored instrument displa
 ## SPC Credentials
 
 * When more than one SPC Credential matches input data, the browser must choose the authenticator in the order of the input SPC Credential Identifiers.
+
+Q: Does it mean that if more than one SPC Cred IDs are valid on the given device/browser then the browser picks first on the list and does not present selection?
+
 * When no SPC Credential matches input data, the protocol should terminate without any user experience to allow for seamless fallback behaviors.
 * If the protocol supports more than one instrument per authenticator (e.g., within the same SPC Credential), then each instrument must be uniquely addressable and have unique display information.
+
+Q: I need clarification here, please. If two instruments can be bound to the single SPC Credential then how SPC is going to be invoked? By SPC Credenial ID of Instrument ID?
 
 ### Lifecycle Management
 
@@ -83,6 +105,7 @@ authorization it can be validated against relying-party stored instrument displa
 ## SPC Assertions
 
 * The SPC Assertion must include at least: a merchant origin/identifier, amount and currency, transaction id.
+* The SPC Assertion must include the cryptographic nonce / challange if provided within the SPC request.
 * The SPC Assertion must also include information about the user's journey. This information may be part of the authenticator's own assertion. For example, the assertion must indicate whether the user completed the transaction without a user presence check.
 * The SPC Assertion must also include information about which entity displayed the transaction confirmation user experience (browser, OS, or authenticator).
 * The SPC Assertion must include a signature over that data (based on the associated authenticator). This signature may be validated by the Relying Party or any other party with the appropriate key.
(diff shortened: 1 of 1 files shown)


requirements.md
```diff
@@ -16,29 +16,46 @@ See also: [SPC Scope](scope.md) for definitions and more information.
 ## Web Context Support
 
 * It must possible to invoke SPC from an iframe. See [issue 68](https://github.com/w3c/secure-payment-confirmation/issues/68).
+
+Q: I understand that woudl also require PR API to be used from within the IFrame. Is that right?
```
stephenmcgruer, line 20: If SPC is ultimately spec'd in terms of the PR API, yes. The explainer/experimental version of SPC was done that way (`new PaymentRequest([{supportedMethods: 'secure-payment-confirmation', ...}], ...).show()`), but this doesn't mean we _have_ to write the 'real' spec the same way (though some of us are biased in that direction ;)).

Do you have a concern regarding 'PR API to be used from within the IFrame'?

requirements.md
```diff
@@ -16,29 +16,46 @@ See also: [SPC Scope](scope.md) for definitions and more information.
...
+
 * It must possible to invoke SPC from within Payment Request API.
+
+Q: Shouldn't that be rather "It must be possible to invoke SPC via Payment Request API"? I understand the intention is to use the PR API as a way to invoke the SCP.
+
 * It must possible to invoke SPC from within a Payment Handler.
 
+Q: Does is mean the Payment Handler API woudl need to be extended by some new capabilities to trigger the SPC from within the PH code?
```
stephenmcgruer, line 28: Possibly, if SPC ends up using the PR API as a basis. One cannot invoke PR from within a PaymentHandler (because 
 PaymentRequest.show() cannot be nested), so some other API would be needed in that case.

Straw-person _example_ of how this could work: it could be possible both to trigger SPC via `new PaymentRequest(...).show()` **and** adding a new method to [PaymentRequestEvent](https://www.w3.org/TR/payment-handler/#the-paymentrequestevent) called `invokeSPC()`.

requirements.md
```diff
@@ -16,29 +16,46 @@ See also: [SPC Scope](scope.md) for definitions and more information.
...
+
 ## Enrollment
 
 * It must be possible to enroll an SPC Credential outside of a transaction.
 * It must be possible to enroll an SPC Credential during a transaction. This enrollment should not prevent timely completion of the transaction.
+
+Q: Does that also imply that the enrolment during the transaction coudl somehow produce the SPC Attestation? Such capability woudl help to solve one of the difficult problems of getting conusmer authenticated for the first transaction.
```
stephenmcgruer, line 35: I don't think this is possible due to the trust model (not due to any technical reason!). The relying party has to already have authenticated the user in some form in order to enroll them, as otherwise how does it know they are who they say they are! (Stolen credit card, etc). So at the point where one can trust the user to be enrolled, they have likely already logged into their bank, or performed an OTP process, etc.

If I am wrong for some reason, we could in theory extend the enrollment process to also sign something.

stephenmcgruer (commented):
I've left a few comments on parts I think I can answer concretely, but there are a lot of other great questions here! Hopefully others from the WG will weigh in because in some cases I have similar questions :)

requirements.md
```diff
@@ -73,16 +90,22 @@ authorization it can be validated against relying-party stored instrument displa
 ## SPC Credentials
 
 * When more than one SPC Credential matches input data, the browser must choose the authenticator in the order of the input SPC Credential Identifiers.
+
+Q: Does it mean that if more than one SPC Cred IDs are valid on the given device/browser then the browser picks first on the list and does not present selection?
```
stephenmcgruer, line 94: I think this is an interesting question, but I'm not sure what giving a choice here to the user would do. It might be useful for us to think through where the multiple matching credentials may have come from (i.e. what are the journeys that led there), and what difference it makes to the user which one they select.

requirements.md
```diff
@@ -91,6 +114,8 @@ authorization it can be validated against relying-party stored instrument displa
 
 * For privacy, the protocol must enable the user to require that two payment instruments NOT be associated with the same authenticator.
 
+Q: Please clarify.
```
stephenmcgruer, line 117: I think this bullet point is meant to say 'with the same credential' rather than authenticator, but I would also still like to understand the privacy concern in that case better as well :D

stephenmcgruer (commented):
(A few more comments, now on things I don't have concrete answers for :D)

requirements.md
```diff
@@ -16,29 +16,46 @@ See also: [SPC Scope](scope.md) for definitions and more information.
 ## Web Context Support
 
 * It must possible to invoke SPC from an iframe. See [issue 68](https://github.com/w3c/secure-payment-confirmation/issues/68).
+
+Q: I understand that woudl also require PR API to be used from within the IFrame. Is that right?
+
 * It must possible to invoke SPC from within Payment Request API.
+
+Q: Shouldn't that be rather "It must be possible to invoke SPC via Payment Request API"? I understand the intention is to use the PR API as a way to invoke the SCP.
```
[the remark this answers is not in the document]
ianbjacobs: One idea had been to create something like an authenticate() method. In a world where show() is used to get instrument information, it could be followed by authenticate(). Or, one could call authenticate() without calling show() if the instrument is already known.

requirements.md
ianbjacobs: Good question. I am going to clarify this point to be: " It must be possible to enroll an SPC Credential following a prior (non-SPC) authentication during a transaction." 

I don't think the expectation was that SPC enrollment would take place without prior authentication. If that is a requirement, let's add it.

requirements.md
```diff
@@ -16,29 +16,46 @@ See also: [SPC Scope](scope.md) for definitions and more information.
...
 ## Enrollment
 
 * It must be possible to enroll an SPC Credential outside of a transaction.
 * It must be possible to enroll an SPC Credential during a transaction. This enrollment should not prevent timely completion of the transaction.
+
+Q: Does that also imply that the enrolment during the transaction coudl somehow produce the SPC Attestation? Such capability woudl help to solve one of the difficult problems of getting conusmer authenticated for the first transaction.
+
 * It must be possible to enroll an SPC Credential from code on a merchant site.
+
+Q: Can you elaborate more what's the use-case? My assumption was that the RP i.e. issuer or wallet would enroll the SPC Credential. The merchant or PSP woudl be leveraging the enrolled credential by triggering the SPC auth flow.
```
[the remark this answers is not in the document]
ianbjacobs: I agree with the point. I am going to change the bullets to be:

* It is not a requirement that the relying party be able to enroll an SPC Credential in a third-party context.
* It must be possible to enroll an SPC Credential from a payment handler.

requirements.md
ianbjacobs: I am going to remove this requirement and instead link to issue 69 for discussion.

requirements.md
```diff
@@ -73,16 +90,22 @@ authorization it can be validated against relying-party stored instrument displa
...
 
 * When more than one SPC Credential matches input data, the browser must choose the authenticator in the order of the input SPC Credential Identifiers.
+
+Q: Does it mean that if more than one SPC Cred IDs are valid on the given device/browser then the browser picks first on the list and does not present selection?
+
 * When no SPC Credential matches input data, the protocol should terminate without any user experience to allow for seamless fallback behaviors.
 * If the protocol supports more than one instrument per authenticator (e.g., within the same SPC Credential), then each instrument must be uniquely addressable and have unique display information.
 
+Q: I need clarification here, please. If two instruments can be bound to the single SPC Credential then how SPC is going to be invoked? By SPC Credenial ID of Instrument ID?
+
```
ianbjacobs, line 100: I can imagine a world where we structure the data so that there are N instruments per authenticator credential. In this case, the SPC Credential ID would reference a specific instrument associated with the SPC Credential. 

In short: it seems like we should discuss the use case of N instruments per authenticator credential and see if we want to support it. If we do, then I think we can still talk about "SPC Credential ID" and the browser will figure it out.

requirements.md
tblachowicz: Not really. I just wanted to be clear that the invocation of SPC auth from within the IFrame would be the same as from the main context. I think we're on the same page.

requirements.md
ianbjacobs, line 117: I am removing this one for now. And I closed issue 43 which I think was poorly expressed in any case.

Here's the goal I have in mind:

* If a person has two instruments and has done what's necessary so that merchants paid with those instruments do not have enough information to correlate them to the user, then the web protocol itself should not provide information to enable correlation.

requirements.md
tblachowicz: Thanks. So the actual authentication would need to happen prior the new `PaymentRequest(...).show()`. Previously we discussed the concept of a fallback mechanism that would allow RP to leverage the 1PC context dialog to perform the authand then enroll the SPC credential following the successful authentication.

requirements.md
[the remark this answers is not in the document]
tblachowicz: @ianbjacobs: interesting idea. I'm not very keen on using the payment method identifier literal to trigger the SPC, so perhaps the explicit method would be an interesting alternative.

requirements.md
tblachowicz: Make sense. I think the relationship between the instruments and credentials needs to be clearly defined because it's reflected on the API.

requirements.md
tblachowicz: If I understood correctly it'd be contradictory with the previous point (https://github.com/w3c/secure-payment-confirmation/pull/71#discussion_r632531886)

requirements.md
ianbjacobs: Hearing "same page" and having updated the doc, I'm going to close this part of the conversation.

requirements.md
ianbjacobs: i've added a note to the requirements document that this is likely to require updates to PH API.

requirements.md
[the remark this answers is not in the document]
ianbjacobs: I'm going to resolve this piece of the discussion since we are moving the document in a way that is independent of PR API and WebAuthn (in terms of capability). The document now says "It must be possible to call SPC from a Web site or a payment handler. "

requirements.md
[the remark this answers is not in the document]
ianbjacobs: Resolving with those two edits.

requirements.md
ianbjacobs: Resolving with that edit in place.

requirements.md
ianbjacobs: I have removed this requirement. The goal was to not reveal through the protocol itself that, for example, the user of a personal card is the same user that has a different business card. So far I don't see the API doing that, but we should keep an eye out for it. I will resolve this conversation since I've removed the (unclear) requirement. I welcome a clearer proposal.

requirements.md
ianbjacobs: Not resolving this yet; see https://lists.w3.org/Archives/Public/public-payments-wg/2021May/0016.html

ianbjacobs: Hi @tblachowicz, there is now a requirement that each instrument/auth credential pair be addressable. I am going to close this part of the conversation.

ianbjacobs:
@tblachowicz do you think we can close this issue?

ianbjacobs:
Closing this issue due to inactivity. We can raise new individual issues as needed.

PR #72, opened 2021-05-14
Comments/questions on requirements.md

stephenmcgruer:
Inspired by Tomasz's create questions/comments in https://github.com/w3c/secure-payment-confirmation/pull/71, here are my own :). I've attempted to only comment on points that Tomasz's doesn't, so I hope there's not any divergence of conversation.

### DIFF
--- a/requirements.md
+++ b/requirements.md
@@ -1,3 +1,7 @@
+Q: Meta-question; are the uses of 'must', 'should' etc here to be interpreted per [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) ?
+
+Q: Meta-question; these requirements encompass a fairly large scope. I believe that we can spec and ship SPC in 'chunks' towards this goal, where each stage might add a few more capabilities meeting these requirements, but not have every single requirement in v1. Do others agree?
+
 # Secure Payment Confirmation: Requirements and Design Considerations
 
 Status: This is a draft document without consensus.
@@ -57,6 +61,9 @@ See also: [SPC Scope](scope.md) for use cases as well as the [issues list](https
 #### Instrument Information
 
 * Enrollment of an instrument must include display information for it.
+
+Q: Should this be a 'must' if we're allowing for per-RP credentials (i.e. as opposed to requiring per-payment-instrument)? I could see it as an optional include in that case.
+
 * Because instrument display information is available to the relying party (e.g., provided by the relying party itself, the merchant, or some other party), it is not a requirement that this information be stored in the browser as part of the [SPC Credential](#dfn-spc-credential).
 * The relying party should be able to update the instrument information of an enrolled [SPC Credential](#dfn-spc-credential) (e.g., for new card art).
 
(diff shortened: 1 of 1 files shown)


requirements.md
```diff
@@ -30,13 +34,15 @@ See also: [SPC Scope](scope.md) for definitions and more information.
...
+
+Q: Should this be a 'must' if we're allowing for per-RP credentials (i.e. as opposed to requiring per-payment-instrument)? I could see it as an optional include in that case.
+
 * Because instrument display information is available to the relying party (e.g., provided by the relying party itself, the merchant, or some other party), it is not a requirement that this information be stored in the browser as part of the SPC Credential.
 
 ## Transaction Confirmation User Experience
 
 * Each browser must natively support a transaction confirmation user experience.
 * Although we anticipate that in most cases the browser will render the transaction confirmation user experience, the protocol must support rendering by other entities (e.g., the operating system or authenticator).
-* See [issue 48](https://github.com/w3c/secure-payment-confirmation/issues/48) on merchant information display.
```
stephenmcgruer, line 39: This is the same bullet as two-down, I think.

requirements.md
```diff
@@ -1,3 +1,7 @@
+Q: Meta-question; are the uses of 'must', 'should' etc here to be interpreted per [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) ?
```
ianbjacobs, line 1: Yes. I'll add a note to that effect.

requirements.md
```diff
@@ -1,3 +1,7 @@
...
+
+Q: Meta-question; these requirements encompass a fairly large scope. I believe that we can spec and ship SPC in 'chunks' towards this goal, where each stage might add a few more capabilities meeting these requirements, but not have every single requirement in v1. Do others agree?
+
```
ianbjacobs, line 4: +1. This list is currently unprioritized.

requirements.md
```diff
@@ -30,13 +34,15 @@ See also: [SPC Scope](scope.md) for definitions and more information.
 ### Instrument Information
 
 * Enrollment must include display information for the associated instrument.
+
+Q: Should this be a 'must' if we're allowing for per-RP credentials (i.e. as opposed to requiring per-payment-instrument)? I could see it as an optional include in that case.
```
ianbjacobs, line 38: Interesting comment. I can imagine multiple use cases:
 * One authentication credential for one instrument.
 * One authentication credential for multiple instruments, where each instrument is independently addressable.
 * One authentication credential per user profile, where individual instruments are not independently addressable. (In which case I agree with your point.)

Those are independent of the ability to have multiple authentication credentials (e.g., phone, desktop) for the same instrument.

Do the above three make sense? Are there others? Is there a single model that enables the above three?

requirements.md
ianbjacobs: Agreed. Removed the dup and closing here.

requirements.md
ianbjacobs: Noted this status in the doc (unprioritized) and closing here.

requirements.md
ianbjacobs: See concrete proposal for 1-1 cardinality authentication-to-instrument:
 https://lists.w3.org/Archives/Public/public-payments-wg/2021May/0016.html

This proposal enables "simultaneous enrollment of N instruments for one authentication credential", resulting in N "per-instrument" credentials. It does not support "per-RP credentials." It's not clear to me what that would mean in practice. 

It is not clear that "RP" is the right level of granularity. For example, a person might have multiple identities with the same RP. I also think that "profile-at-PR" is not the right granularity since one might have multiple accounts for a given identity.

stephenmcgruer:
(Rebased past resolved comments and recent changes. No actual change in comments/questions other than moving them as necessary).

requirements.md
ianbjacobs: @stephenmcgruer, I think we can close this thread:
* I do not think we want "per-RP" credentials (e.g., a user might have multiple accounts with an RP).
* There is now a pull request to support dynamic binding to an instrument at authentication time.
* There is a requirement that each instrument/auth credential pair (SPC Credential) be individually addressable.

If you agree, I think you can close the pull request. :)

Thanks!

stephenmcgruer: I agree, but for completeness will note:

> I do not think we want "per-RP" credentials (e.g., a user might have multiple accounts with an RP).

At the risk of being a broken-record, 'per-RP' was my lazy way of saying 'cardinality should be up to the RP', which I communicated poorly. (Though I believe webauthn folks also often say 'per-RP' when they really mean 'per account or other state concept that the RP decides on' :D)


I'm rebasing atm, will close the PR if post-rebase everything seems good.

stephenmcgruer:
After rebase and reading the current version of requirements.md, I agree that this PR is now obsolete.

PR #73, opened 2021-05-18
Add user-activation note to requirements.md

stephenmcgruer:


stephenmcgruer:
So, I wrote this deliberately taking a viewpoint (of requiring activation to call SPC; approximately the user has to click a button before seeing any browser UX). This is in-line with `PaymentRequest.show()` today. Gerhard may disagree with me, but good to have that discussion :D

Note that in researching this, I found out that WebAuthn today **does not require user activation** by spec o.O ([though WebKit made it a requirement for most sites](https://github.com/w3c/webauthn/issues/1293#issuecomment-829751750)). That was very surprising to me, but it seems that it has been hotly debated - https://github.com/w3c/webauthn/issues/1293 - and is still ongoing as of two weeks ago.

ianbjacobs:
This requirement makes sense to me. In practice, some user gesture should happen, including:
* entering instrument information into a form
* selecting a card-on-file
* selecting a payment handler
* selecting an instrument from a list of instruments (e.g., SRC display)

My understanding is that all of our low-friction flows take place after some instrument identification action.

rlin1 (approved):

rsolomakhin:
@stephenmcgruer - Should we require a user gesture in enrollment, too?

ianbjacobs:
@rsolomakhin, one thing occurs to me: enrollment may be "less sensitive" than authentication because there is no payment. That does not imply "no gesture is needed" but it is one salient difference.

stephenmcgruer:
> @stephenmcgruer - Should we require a user gesture in enrollment, too?

Given we're performing essentially the same action as webauthn here, I'd yield to them for now. We can always make it a `NOTE` in the spec that we'll follow whatever they decide.

PR #74, opened 2021-05-24
Question regarding feature detection

jcemer-stripe:


ianbjacobs:
Hi @jcemer-stripe,

This comment feels like a detailed API suggestion rather than a general requirement. Is there a higher level requirement that you have in mind (that refines the current point about programmatic testing)? Thanks!

ianbjacobs:
Now that we have published a FPWD with a particular approach, I'm going to close this. We can return to the topic as needed.

PR #78, opened 2021-05-28
Allow for dynamic binding of instrument

ianbjacobs:
Per discussion on 24 May, this edit seeks to support dynamic binding of instruments (at auth time instead of static binding at enrollment time). Also moved around and regrouped existing requirements (improved organization).

### DIFF
--- a/requirements.md
+++ b/requirements.md
@@ -18,10 +18,10 @@ See also: [SPC Scope](scope.md) for use cases as well as the [issues list](https
 : A mechanism used to transfer value from a payer to a payee.
 
 **SPC Credential** <a name="dfn-spc-credential"></a>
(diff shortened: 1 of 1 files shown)


ianbjacobs:
@rsolomakhin and @stephenmcgruer, this pull request intends to support the dynamic binding of instrument information. It's a little noisier than you might expect because I did some reorganization at the same time. 

The key changes:

* This is gone: "Enrollment of an instrument must include display information for it."
* This is new: "The Relying Party must not be required to provide definitive information about a specific instrument at enrollment time. In other words, the API supports dynamic binding to a specific instrument at authentication time."

@stephenmcgruer, it would help to add a use case to scope.md that illustrates the dynamic binding. Would you be able to write that up? Thank you!

rsolomakhin (approved):
This update looks good to me. Thank you!

stephenmcgruer:
LGTM, thanks!

PR #88, opened 2021-07-13
[Spec] First draft version of spec

stephenmcgruer:
This version of the spec is an early-draft, laying out one possible
API shape for SPC. It should not be considered final, and exists to give
spec discussions a concrete basis to debate on.

Also-by: Rouslan Solomakhin <rouslan@chromium.org>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1,4 +1,3 @@
-Date:   Wed Jul 7 11:24:50 2021 -0400
 <pre class="metadata">
 Title: Secure Payment Confirmation
 Shortname: secure-payment-confirmation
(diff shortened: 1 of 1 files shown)


ianbjacobs (commented):
@stephenmcgruer and @rsolomakhin,

First thank you!! Do you think we should merge this initial draft then start sending individual pull requests?

Ian

stephenmcgruer:
I would suggest we do, just because its easier to review individual separate pull requests (or issues, if folks have concerns but no concrete fix is clear yet) instead of comments on this larger PR. We can also start to hot-link issues into the spec where there are ongoing discussion points.

ianbjacobs (approved):

rsolomakhin:
> should merge?

Agreed.

Issue #83, opened 2021-06-28
Popup  UI handling of origin - relying party vs. merchant


weksler:
The SPC popup currently shows the relying party only (e.g. adyen.com), which might be confusing to the end user, because they are transacting with the merchant (e.g. airbnb.com). Might there be a way to include that information in the SPC popup _as well_, e.g say something like ".... adyen.com on behalf of airbnb.com..." ?

stephenmcgruer:
I believe the statement 'The SPC popup currently shows the relying party only' is incorrect, but I think I know where it is coming from.

## What we show currently

For Chrome, there are two UIs shown during authentication - the SPC transaction details dialog, and the platform-drawn WebAuthn dialog. The former currently shows the top-level domain as the 'merchant', which is automatically derived within the browser (i.e. we don't ask the caller of SPC to pass it in). The WebAuthn dialog, on the other hand, will show the relying party of the credential (e.g. the bank), as that is whom the user is authenticating to.

In the following screenshots ([demo](https://rsolomakhin.github.io/pr/spc-iframe/)):

* The 'merchant' is https://rsolomakhin.github.io/ (top-level origin)
* The 'payment service provider' is https://emerald-spiced-psychology.glitch.me/ (caller of SPC, inside an iframe)
* The 'bank' is https://lumbar-brick-soup.glitch.me/

![Screenshot 2021-06-28 154621](https://user-images.githubusercontent.com/153264/123695154-35553480-d828-11eb-8678-0fd3f095a9ae.jpg)

![Screenshot 2021-06-28 154641](https://user-images.githubusercontent.com/153264/123695434-8a914600-d828-11eb-9cae-f22fc784514f.jpg)

Notably, the payment service provider in this example is never shown to the user, but it is included in the signed cryptogram (it's the `origin` in the collected client data, IIRC). Whether it should be shown in the UI is a fair question (likely comes down to what the user agent thinks is good + what might be required for e.g. SCA compliance).

## The catch

The catch is that SPC, as currently written, **assumes the user is on the merchants web-page when it is called** (though the call itself may happen in a cross-origin subframe). That only works if iframes are being used, and what I suspect @weksler is referring to is the behavior seen when one uses a redirect model (i.e. redirects away from merchant.com to either psp.com or bank.com).

In that case, the UI will look weird, because we'll show the top-level frame as the merchant (which it clearly won't be), and the cryptogram will also contain no assertion about who the merchant is. For example, in the psp.com redirect example:

[...]

PR #89, opened 2021-07-16
[Spec] Add payeeOrigin

stephenmcgruer:
Fixes https://github.com/w3c/secure-payment-confirmation/issues/83

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -419,6 +419,7 @@ The [=standardized payment method identifier=] for this specification is
         required FrozenArray<BufferSource> credentialIds;
         required PaymentCredentialInstrument instrument;
         unsigned long timeout;
+        required USVString payeeOrigin;
         AuthenticationExtensionsClientInputs extensions;
     };
 </xmp>
@@ -438,6 +439,9 @@ members:
 
     :  <dfn>timeout</dfn> member
     :: The number of milliseconds before the request to sign the transaction details times out. At most 1 hour.
+
+    :  <dfn>payeeOrigin</dfn> member
+    :: The origin of the payee that this SPC call is for (e.g., the merchant).
 
     :  <dfn>extensions</dfn> member
     :: Any [=WebAuthn extensions=] that should be used for the passed credential(s). The caller does not need to specify the [[#sctn-payment-extension-enrollment| payment extension]]; it is added automatically.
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
I'm curious on folks' thoughts on whether to make this field required, or whether to auto-compute it if empty (to `topOrigin`), or to leave it empty if not set.

Given the UX is likely showing it, I think we have to either do required or auto-computed.

rsolomakhin:
Requiring the field should make the algorithms simpler without putting too much burden on the web developers. That's my personal preference, but I'm happy to hear other opinions.

ianbjacobs:
@jcemer, @apouillard, @a-akimov, @mweksler, @bdewater, @Goosth, any thoughts on this?
Ian

Goosth:
We need to consider the different form-factors. E.g. what if the merchant does not support SPC, but the Issuer/Account Provider does, and wants to use it in the Account Provider context. At first glance I would say
* It makes sense to support some form of topOrigin/callingparty field
* Should be auto-filled by the browser
* Should be included in signature.

apouillard:
I'd rather make this merchantOrigin a required field. However, as the payee is not always a merchant, maybe should we think about a more generic name for this field...

SensibleWood:
> I'd rather make this merchantOrigin a required field. However, as the payee is not always a merchant, maybe should we think about a more generic name for this field...

+1 - it could be a PISP, for example

stephenmcgruer:
Renaming is fine to me. `payeeOrigin`? `onBehalfOf`?

There are probably two concepts to keep in mind while we think about this:

1. What is shown in the auth-time UX, both for regulatory reasons and for user understanding.
  * This sub-divides into what a UA may want to show, and what a UA has to show (for regulatory reasons). 
2. What concept ends up signed in the cryptogram.

rsolomakhin:
> What concept ends up signed in the cryptogram.

The `payeeOrigin` proposal sounds good.

stephenmcgruer:
PR updated to use `payeeOrigin`, and to make it `required` in SPC auth.

ianbjacobs (approved):

rsolomakhin (approved):
LGTM, thank you!

PR #90, opened 2021-07-21
Editorial review of section 1

ianbjacobs:
Looking forward to discussion of these proposed changes.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -12,7 +12,10 @@ Level: 1
 URL: https://w3c.github.io/secure-payment-confirmation
 Editor: Rouslan Solomakhin, Google https://www.google.com/, rouslan@chromium.org
 Editor: Stephen McGruer, Google https://www.google.com/, smcgruer@chromium.org
-Abstract: Secure Payment Confirmation (SPC) is a Web API to support streamlined authentication during a payment transaction. It is designed to scale authentication across merchants, to be used within a wide range of authentication protocols, and to produce cryptographic evidence that the user has confirmed transaction details.
+Abstract: Secure Payment Confirmation (SPC) is a Web API to support streamlined
+  authentication during a payment transaction. It is designed to scale authentication
+  across merchants, to be used within a wide range of authentication protocols, and
+  to produce cryptographic evidence that the user has confirmed transaction details.
 Complain About: missing-example-ids true
 Markup Shorthands: markdown yes
 </pre>
--- a/spec.bs
+++ b/spec.bs
@@ -12,8 +12,7 @@ Level: 1
 URL: https://w3c.github.io/secure-payment-confirmation
 Editor: Rouslan Solomakhin, Google https://www.google.com/, rouslan@chromium.org
 Editor: Stephen McGruer, Google https://www.google.com/, smcgruer@chromium.org
-Abstract: This specification describes data structures, formats, algorithms, and
-  processing models to facilitate [[webauthn-3]] based payments on the Web.
+Abstract: Secure Payment Confirmation (SPC) is a Web API to support streamlined authentication during a payment transaction. It is designed to scale authentication across merchants, to be used within a wide range of authentication protocols, and to produce cryptographic evidence that the user has confirmed transaction details.
 Complain About: missing-example-ids true
 Markup Shorthands: markdown yes
 </pre>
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -56,18 +55,17 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
...
+
+Similarly to [[webauthn-3]], this specification defines two related
+processes involving a user. The first is [[#sctn-enrollment]], where a
+relationship is created between the user and the [=Relying
+Party=]. The second is [[#sctn-authentication]], where the user
+responds to a challenge from the [=Relying Party=] (possibly via an
+intermediary payment service provider) to consent to a specific
+payment. An important feature of Secure Payment Confirmation is that
+the merchant (or another entity) may initiate the authentication
+ceremony on the [=Relying Party's=] behalf and with their permission.
```
stephenmcgruer, line 68: Alas, 'with their permission' is not true. Nothing stops merchants from sharing 3p credentials, at which point anybody may initiate the ceremony. It's not **useful** for the entity that is initiating it, but they technically can.

spec.bs
```diff
@@ -12,8 +12,7 @@ Level: 1
 URL: https://w3c.github.io/secure-payment-confirmation
 Editor: Rouslan Solomakhin, Google https://www.google.com/, rouslan@chromium.org
 Editor: Stephen McGruer, Google https://www.google.com/, smcgruer@chromium.org
-Abstract: This specification describes data structures, formats, algorithms, and
-  processing models to facilitate [[webauthn-3]] based payments on the Web.
+Abstract: Secure Payment Confirmation (SPC) is a Web API to support streamlined authentication during a payment transaction. It is designed to scale authentication across merchants, to be used within a wide range of authentication protocols, and to produce cryptographic evidence that the user has confirmed transaction details.
```
stephenmcgruer, line 15:
```suggestion
Abstract: Secure Payment Confirmation (SPC) is a Web API to support streamlined
  authentication during a payment transaction. It is designed to scale authentication
  across merchants, to be used within a wide range of authentication protocols, and
  to produce cryptographic evidence that the user has confirmed transaction details.
```

stephenmcgruer (approved):
LGTM! (Some formatting nits about not wrapping consistently, but minor enough that I won't complain :P)

spec.bs
ianbjacobs: The current text is: "An important concept in Secure Payment Confirmation is that with the permission of the Relying Party...."

I preserved "with their permission" but have no problem dropping it.

Having said that, "permission" is still relevant even if it might not be the ideal term. The RP would have to share the credential ID, which is a form of permission. 

Are you ok to simply drop "and with their permission"?

stephenmcgruer: Ahaha, touche. I apologise, and send the original criticism to myself instead.

We can drop 'and with their permission'

spec.bs
```diff
@@ -76,77 +77,94 @@ allow the API to be called in payment contexts.
...
+1. There is no specification for this, so each bank is likely to have
+    to devise its own format for how payment-specific information should be formatted and encoded in the challenge, complicating deployment and increasing fragmentation.
+1. Regulations may require evidence that the user was shown and agreed to the payment-specific information. Plain [[webauthn-3]] does not provide for this display: there is no specified UX associated with information stored in the `challenge` field.
+
+These limitations motivate the following Secure Payment Confirmation behaviors:
+
+1. The `challenge` field is only used to defeat replay attacks, as with plain
+    [[webauthn-3]].
+1. SPC specifies a format for payment-specific information.
+   This will enable development
```
stephenmcgruer, line 116:
```suggestion
    This will enable development
```

ianbjacobs:
@stephenmcgruer, regarding "(Some formatting nits about not wrapping consistently, but minor enough that I won't complain :P)"...is there a "tidy" equivalent that I can use with bikeshed to get consistent formatting?

stephenmcgruer:
> @stephenmcgruer, regarding "(Some formatting nits about not wrapping consistently, but minor enough that I won't complain :P)"...is there a "tidy" equivalent that I can use with bikeshed to get consistent formatting?

Alas, not as far as I know. Don't worry too much about it, as long as it builds its good :)

ianbjacobs:
Aha, there are bikeshed conventions I don't know about and I need to make fixes. Got it.

stephenmcgruer:
Btw if you have an actual checkout, you can run `make remote` to build it without any dependencies other than an internet connection. But of course one might be using the GitHub UI which does make it harder!

ianbjacobs:
@stephenmcgruer, thanks for reviewing the PR!

@rsolomakhin, I'll wait for your ok before merging.

rsolomakhin:
Looks good, thank you!

rsolomakhin (approved):

PR #95, opened 2021-07-27
Remove custom CollectedClientData handling

adrianhopebailie:
1. Don't set type to payment.create

Given the goal of aligning more closely with WebAuthn the use of `payment.create` instead of `webauthn.create` when creating a new credential seems arbitrary. The data scheme a is the same so there is no need to detect the type.

Also, there is no Auhtenticator extenion processing so no need for authenticators to treat this client data in a special manner.

2. Remove note re getClientExtensionResults

Since the extension defines no "Client extension output" this would always return an empty result (or am I misreading the spec?)

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -610,10 +610,6 @@ directly; for authentication the extension can only be accessed via
          consume a [=transient activation=] here, if we felt the permission policy
          is not sufficient.
 
-    2. In step 13, set the {{CollectedClientData/type}} to "`payment.create`".
-
-        <div class="note">**TODO**: Can we and/or should we rely on `getClientExtensionResults()` instead?</div>
-
 :  Client extension processing ([=authentication extension|authentication=])
 :: When [[webauthn-3#sctn-getAssertion|making an assertion]]:
 


davordavidovic:
I second this and also want to suggest the same for `payment.get` during authentication. A couple of reasons for this based on our use case:

1.  We (and probably most other relying parties) use an external webauthn library for validating the authentication/registration response server-side. Making a change to the accepted credential `type` values would probably mean maintaining our own version of that library (at least at first) which creates a lot of overhead.
2. As far as I understand the `type` field does not add any security value. The `clientDataJSON` already contains the `payment` object, indicating an SPC response. Please correct me if I'm wrong on this.

rsolomakhin:
After double-checking with our teams internally, we have come to the conclusion that the `"payment.get"` type for authentication must stay as `"payment.get"` because breaking the server-side libraries is intentional. Many vulnerabilities in the paste have been caused by multiple parties in a protocol having a misunderstanding about exactly what is going on. As such, a WebAuthn validation library should reject "did user authenticate?" query when the user's intent was "authorize a transaction" instead. One salient difference in particular is that 3rd parties (e.g., PSPs or merchants) can generate signatures for payments on behalf of RPs (e.g., banks), whereas vanilla WebAuthn requires that only RP (e.g., bank) can generate a signature for itself. 

Following the same logic, enrolling a credential should have `clientDataJson.type` of `"webauthn.get"` only if the properties of the enrollment correspond to that of vanilla WebAuthn. In prior versions of SPC, a credential could be enrolled from an iframe and enrollment had to be confirmed in browser UX. That was different enough from WebAuthn, where a credential could be enrolled only from a top-level context and there was no browser UX. Now we are hearing that WebAuthn is likely to allow enrollment of credentials from iframes (similar to SPC) and SPC has plans to remove the enrollment UI (similar to WebAuthn). These two changes would remove many of the differences between the SPC and WebAuthn APIs when it comes to enrollment. Therefore, it should be OK to keep `clientDataJson.type` as `"webauthn.create"` even in the case of SPC.

Issue #81, opened 2021-06-24
Make it easy to determine whether a user agent supports SPC


ianbjacobs:
Based on conversations with developers, there is a clear desire for an easy mechanism to know whether a browser/user agent supports SPC.

(Related: Issue #65.)

stephenmcgruer:
To add some context, we've heard from multiple developers experimenting with the API that the 'standard' way of detecting support for a PaymentMethod (as SPC is currently implemented for the Origin Trial) is too clunky:

```
const canUseSPC = await (new PaymentRequest([
  {
    supportedMethods: 'secure-payment-confirmation',
    data: {
      credentialIds: [Uint8Array.from(atob('Q1J4AwSWD4Dx6q1DTo0MB21XDAV76'), c => c.charCodeAt(0))],
      challenge: ...,
      timeout: ...,
    },
  }], {total: {label: 'total', amount: {currency: 'USD', value: '20.00'}}})).canMakePayment()
```

The approach above conflates 'supported in this browser', and 'this PaymentRequest is valid'. Developers often want to answer the former question long before the latter (to decide what flow they will take in their code), and at that point are likely to not even have the credentialIds, the challenge, or maybe even the amount being charged.

ianbjacobs:
@stephenmcgruer, do you have suggestions for how to accomplish this (and outside of Payment Request)?

stephenmcgruer:
> @stephenmcgruer, do you have suggestions for how to accomplish this (and outside of Payment Request)?

Yeah, unfortunately it's a tricky case. If we keep `PaymentCredential` around as an IDL type, it's could be as simple as checking for that existence. But if we succeed in getting closer to WebAuthn, that would be replaced with `PublicKeyCredential` anyway so one wouldn't be able to feature-detect via it.

We could take the same route as `CSS.supports` or `HTMLMediaElement.canPlayType`, which take in descriptors of the feature and return a roughly-boolean answer (`canPlayType` actually returns 'probably', 'maybe', or an empty-string for definitely-not, I think!). These tend to be viewed as last resorts in API space, as they tend to lead to confusing situations (`PaymentRequest.isSupportedMethod('secure-payment-confirmation')` returned true, but when I created a PaymentRequest it failed, why?! <-- could happen if the PaymentRequest itself is invalid, etc).

stephenmcgruer:
Coming back to this; in the way the spec is evolving, `PaymentCredential` will not be a thing. So checking for that is not sufficient.

[...]

Issue #65, opened 2021-05-06
Should SPC be invokable outside of a PaymentRequest flow?


stephenmcgruer:
Ian and I were chatting today and discovered that we had very different mental models about when SPC authentication is invokable; it seems worthwhile to hash these out and come to a conclusion that we can document (as part of the Scope document?).

Some casual definitions:
  - A payment context: a state in which the user views themself as in the process of a financial transaction (often referred to as being in a 'payment flow'). Divided into:
    - An informal payment context: a payment context in which the flow is not visible to the browser (e.g. a checkout flow implemented with generic web APIs).
    - A web-api payment context: a payment context in which the flow is happening via payment-related web APIs (e.g. **inside** of `PaymentRequest.show()`, such as a web-based `PaymentHandler`) and as such is visible to the browser.

Given these definitions, I believe that SPC authentication should be invokable in **both** informal and web-api payment contexts. Do others agree?

Note that if SPC is invokable in informal payment contexts, it is invokable anywhere on the web as the browser cannot tell the difference. It should always require a user gesture, imo, and the browser UX that will be shown should always state a payment context.

stephenmcgruer:
For completeness: SPC as described in the explainer is actually an informal payment context, not a web-api context. Although explainer-SPC **uses** the PaymentRequest API, it is not part of a PaymentRequest flow; it's just piggy-backing on the existing infrastructure. The explainer version of SPC actually can't be used inside a PaymentRequest flow (e.g. a web PaymentHandler), because PaymentRequests cannot be nested (step 8 of show() - https://w3c.github.io/payment-request/#show-method).

cyberphone:
@stephenmcgruer Since you requested input on a pretty fundamental issue, may I as an external comment on this?  

IMO, SPC is (or should at least be), a browser-built-in payment-handler, activated by the existing `PaymentRequest` API for usage in payment scenarios.  Personally, I would limit the scope even further by making crypto = FIDO.   Going outside of this _already quite challenging scope_ would IMO be unwise (unless it is supported by peer-reviewed, fairly complete specifications, and thoroughly tested code).

Other solutions and scenarios can presumably be supported by existing Web technologies.

In case you want to discuss this further, _please use my email_: anders.rundgren.net@gmail.com.

@rsolomakhin @nickburris @btidor-stripe @jcjones

SensibleWood:
[...]

PR #99, opened 2021-08-10
Note on API shape

ianbjacobs:
New note with background on current shape of API and setting expectation it will change.

Relates to issues #81 and #65 (but does not close them).

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -369,6 +369,19 @@ below and in [[#index-defined-elsewhere]].
 
     <div class="note">TODO: Bikeshed the name.</div>
 
+    <div class="note">This specification describes an API implemented
+    to support a Stripe pilot that began in December 2020. To quickly
+    provide support for that experiment, the Chrome team added SPC
+    support atop existing implementations of the Payment Request and
+    Payment Handler APIs. However, given findings from the pilot and
+    subsequent use case and requirements discussions, there is now
+    general agreement that SPC should be usable independent of Payment
+    Request. To foster more experimentation in a timely fashion we are
+    moving forward with the API in its current form, but we expect
+    (without a concrete timeline) that SPC will move away from its
+    Payment Request origins. For developers, this should improve
+    feature detection, invocation, and other aspects of the API.</div>
+
 # Enrollment # {#sctn-enrollment}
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -369,6 +369,19 @@ below and in [[#index-defined-elsewhere]].
...
+    provide support for that experiment, the Chrome team added SPC
+    support atop existing implementations of the Payment Request and
+    Payment Handler APIs. However, given findings from the pilot and
+    subsequent use case and requirements discussions, there is now
+    general agreement that SPC should be usable independent of Payment
+    Request. To foster more experimentation in a timely fashion we are
+    moving forward with the API in its current form, but we expect
+    (without a concrete timeline) that SPC will move away from its
+    Payment Request origins. For developers, this should improve
+    feature detection, invocation, and other aspects of the API.</div>
```
rsolomakhin, line 383: That's a lot of information... Could it be simplified into this?

```suggestion
    <div class="note">This specification describes an API implemented
    atop existing implementations of the Payment Request. We expect
    (without a concrete timeline) that SPC will move away from its
    Payment Request implementation to align better with WebAuthn.
    For developers, this should improve feature detection, invocation,
    and other aspects of the API.</div>
```

rsolomakhin (changes requested):

spec.bs
ianbjacobs: @rsolomakhin, I think the rationale / history is useful. I don't mind shortening it. Let me edit a bit...

spec.bs
```diff
@@ -369,6 +369,15 @@ below and in [[#index-defined-elsewhere]].
 
     <div class="note">TODO: Bikeshed the name.</div>
 
+    <div class="note">To quickly support an initial SPC experiment,
+    this API was designed atop existing implementations of the Payment
+    Request and Payment Handler APIs. There is now general agreement
+    that SPC should be usable independent of Payment Request.  We
```
rsolomakhin, line 375: Perhaps "There is not general agreement that using SPC independent of Payment Request **should be considered**" or "**should be explored**"? I don't want to make it sound like we have a concrete plan in place.

rsolomakhin (changes requested):

Issue #92, opened 2021-07-22
Conditionally-shown Transaction UX and WebAuthn


stephenmcgruer:
# Conditionally-shown Transaction UX and WebAuthn

Hey folks; I wanted to write up this issue not to suggest any change, but to raise
awareness of a tricky part of SPC and to allow people to digest and perhaps discuss.

This issue meanders along the grey area between purist API and hard reality, so I will
mention Chrome specifically in parts. Generally speaking one can substitute the word
'Chrome' with  'any browser concretely implementing SPC'.

## Preamble - the need for conditionally-shown UX

In the current design, SPC asks the caller (e.g. the merchant) to provide a list of
credentials it has received from the Relying Party (e.g. the bank). The reason it is a list
is that neither the Relying Party or the caller are allowed to know what device a given
credential is for, to protect user privacy. As such, the credentials one has for this user
(who has identified themselves via providing payment instrument information, e.g. a
credit card number) may **or may not** match the current device.

The desired behavior for SPC, as I see it, is that the user should only see the subsequent
Transaction UX (from the browser), if there is a *possibility* that they could successfully
complete SPC - i.e. if one of the credentials in the list matches this device. This means
that the Transaction UX is a *conditionally-shown UX*. When the conditional is false (i.e.
no credentials match), the user should not see the Transaction UX*.

Thus, I hope this establishes the need for conditionally-shown UX. If you disagree here,
stop and leave a comment now because the rest assumes this :D.

\* For privacy, in Chrome the user will likely see another UX instead, that informs
   them their device isn't registered and the site will likely try another method for
   authentication. This prevents an attack in which you can probe for credential
   existence by relying on whether an API like SPC immediately fails or not.

## Conditionally-shown UX: Current approach in Chrome

To enable this conditionally-shown UX today, Chrome does something quite hacky
for SPC; it caches created SPC credentials locally (and in a per-user storage), so that
it can later match them when an SPC call is made. The reason this is necessary is
that WebAuthn does not (in general) allow the browser to discover the existence or lack
thereof of a credential *without the user interacting with their authenticator*.

The browser-local storage works in a world where only Chrome is implementing SPC, and
where users only use one profile per browser (which is generally true), but falls apart quickly
in a multi-browser world. The failure mode is quite straightforward; if a credential is registered
in another browser, Chrome will fail to recognize it as matching this device and will fail an SPC
call that tries to use it. This then has knock-on implications, as it is likely the caller (e.g.
[...]

PR #103, opened 2021-08-16
Suggested clarification and extensions

Goosth:
Using the editing to include suggested additions and clarifications on specific sections of the specification, and to create discussion.
(as recommended by @ianbjacobs).

Initially these are just comments and questions to test the view of the group. I will add suitable text once the concepts or extra examples are agreed to.

***

### :boom: Error: 400 Bad Request :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Aug 16, 2021, 11:30 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fsecure-payment-confirmation%2F8946c61b3a55fd5f4d1858b057500dc58b081cb3%2Fspec.bs&md-warning=not%20ready)

```
Error running preprocessor, returned code: 2.
FATAL ERROR: Line 302 isn't indented enough (needs 1 indent) to be valid Markdown:
"  and both are legitimate?"
 ✘  Did not generate, due to fatal errors
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/secure-payment-confirmation%23103.)._
</details>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -189,6 +189,12 @@ example flows and do not limit the scope of how the API can be used.
 This is the first-time flow, in which a new credential is created and stored by
 an issuing bank.
 
+**TODO**: Suggest we add an extra example, which might be more widely used: When a user navigates to bank.com for internet banking and 1) their account access 
+   is being setup with WebAuthn, or they are offered to extend their WebAuthn access to also include payments.
+  Note that 'activation during shopping' is discouraged in a number of scenario's, and not favioured by merchants. 
+  They would typically not want to hand over control if there is risk of a whole enrollment that would disrupt their experience. 
+  So might be better not to use this as our primary example
+
 1. The user visits `merchant.com`, selects an item to purchase, and proceeds
     to the checkout flow. They enter their payment instrument details, and
     indicate that they wish to pay (e.g., by pressing a "Pay" button).
@@ -212,6 +218,10 @@ an issuing bank.
     credential is created and returned to the issuing bank who stores
     it in their server-side database associated with the user and
     payment instrument(s).
+    
+**TODO**: Define an Agent specific display for Enrollment, indicating that that Payment transactions will be supported (or that this is exclusively for payments)? 
+
+**TODO**: Enhancement to consider: Allow the the user able to select if Payment should be enabled or not (the payment extension) or both?
 
 1. The verification completes; the bank iframe closes and the merchant finishes
     the checkout process for the user.
@@ -229,6 +239,8 @@ const publicKey = {
   rp: {
     name: "Fancy Bank",
   },
+  
+  Q: Should we perhaps require that  rpId (domain) must be specified if this is inside an iFrame? For top level it's probably OK to use the top level domain?
 
   // User:
   user: {
@@ -255,9 +267,12 @@ const publicKey = {
   authenticatorSelection: {
     userVerification: "required"
   },
+  
+  **TODO**: Extend definition (perhaps in future) to make this optional (so to only have userPresence: "required", and thereby support regions outside of europe) and remove friction of having to type in User PIN's?
 
-  timeout: 360000,  // 6 minutes
-
+  timeout: 60000,  // 1 minute
+  Note: (6 minutes might be too long for a merchant checkout scenario... time is critical)
+  
   // Indicate that this is an SPC credential. This is currently required to
   // allow credential creation in an iframe, and so that the browser knows this
   // credential relates to SPC.
@@ -281,6 +296,16 @@ navigator.credentials.create({ publicKey })
 </pre>
 
 </div> <!-- non-normative -->
+
+**TODO**: Sugest we provide clarifying text/expand around the following 
+* Does the extension imply that this credential my ONLY be used for Payment, or does this mean the WebAuthn credential could be used for Login or Payment, 
+  and both are legitimate?
+* Can the payment extension be 'added' once the webAuthn is created? Or must it be set at registration time? and can it be disabled too? 
+   Suggest we define it so that this can be 'enabled' or disabled? So the credential is created for login, and payment is added at a later stage when user agrees. 
+   If not, we may need to explain the 'wokaround' where a new credential will be created and the old one abandoned if user settings/preferences change.
+* How does the Agent display this credential creation differently than the creation of a normal WebAuthn Login credential? 
+  Can we define that 'payment cabable' text must be included for the user to consent to?
+(I can provide this text if we agreee to these concepts and it turns out to be possible)
 
 ### Authentication ### {#sctn-sample-authentication}
 
@@ -338,7 +363,9 @@ const request = new PaymentRequest([{
       icon: "https://fancybank.com/card-art.png",
     }, 
 
-    timeout: 360000,  // 6 minutes
+Q: Sugest we define that displayName is Mandatory and icon is optional?
+  
+    timeout: 60000,  // 1 minute
   }], {
     total: {
       label: "Total",
@@ -347,6 +374,8 @@ const request = new PaymentRequest([{
         value: "5.00",
       },
     },
+    
+  Q: Payment request has a number of other fields as well. I recommend that we  define the exhaustive list of fields acceptable by SPC. Initially, these are should be the full list. Other fields (which will not be displayed) should result in an error when canMakePayment is called.
   });
 
 try {
@@ -358,6 +387,8 @@ try {
 
   // response.data is a PublicKeyCredential, with a clientDataJSON that
   // contains the transaction data for verification by the issuing bank.
+  
+  Q: We need to define the format of Response.data somewhere in this doc, including what fields are present if the user accepts, of if they cancel/timeout.
 
   /* send response.data to the issuing bank for verification */
 } catch (err) {
@@ -391,6 +422,8 @@ below and in [[#index-defined-elsewhere]].
     away from its Payment Request origins. For developers, this should
     improve feature detection, invocation, and other aspects of the
     API.</div>
+    
+    Q: Is the suggestion to create SPC specific methods, or roll everything into WebAuthn, with no custom-UI/logic in the browser agent?
 
 # Enrollment # {#sctn-enrollment}
 
@@ -414,11 +447,13 @@ payment method confirms the transaction with the user and then performs an
 representing the authentication ceremony.
 
 At a high level, authentication for Secure Payment Confirmation is similar to
-[[webauthn-3]], with one major conceptual shift. Secure Payment Confirmation
+[[webauthn-3]], with two major conceptual shifts.
+ * Secure Payment Confirmation
 allows a third-party (e.g., the merchant) to trigger an authentication ceremony
 on behalf of the [=Relying Party=], passing in credentials that it has obtained
 from the Relying Party on some other unspecified channel. See
 [[#sctn-use-case-merchant-authentication]].
+ * Secure Payment Confirmation introduces a payment specific display in the browser agent, which confirms payment information (amount and merchant name) with a user.
 
 ## Payment Method: Secure Payment Confirmation ## {#sctn-payment-method-spc}
 
@@ -464,18 +499,27 @@ members:
 <dl dfn-type="dict-member" dfn-for="SecurePaymentConfirmationRequest">
     :  <dfn>challenge</dfn> member
     :: A random challenge that the relying party generates on the server side to prevent replay attacks.
+    
+    **TODO**: Clarify if this challenge should be set to 32 bytes as for WebAuthn, or can be any string, and the Browser agent will hash it.
 
     :  <dfn>credentialIds</dfn> member
-    :: The list of credential identifiers for the given instrument.
+    :: The list of webauthn credential identifiers for the given instrument.
 
     :  <dfn>instrument</dfn> member
     :: The description of the instrument name and icon to display during enrollment and to be signed along with the transaction details.
 
     :  <dfn>timeout</dfn> member
     :: The number of milliseconds before the request to sign the transaction details times out. At most 1 hour.
+    
+    **TODO**: We need to define how this timeout is calculated. E.g. If the timeout is 60 seconds, and the uer spends 25 seconds on the SPC payment page in the browser, does the WebAuthn request then get initialized with 35 seconds or with 60 seconds? 
+    
+    **TODO**: What do we do if the SPC page is open for more than 60 seconds? Do we cancel without attempting WebAuthn?
 
     :  <dfn>payeeOrigin</dfn> member
     :: The origin of the payee that this SPC call is for (e.g., the merchant).
+    
+    **TODO**: Will this be validated in any way? 
+    **TODO**: Will this be inserted by the browser agent if not provided?
 
     :  <dfn>extensions</dfn> member
     :: Any [=WebAuthn extensions=] that should be used for the passed credential(s). The caller does not need to specify the [[#sctn-payment-extension-enrollment| payment extension]]; it is added automatically.
@@ -533,6 +577,9 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{AuthenticationExtensionsPaymentInputs/instrument}}" → |data|.{{SecurePaymentConfirmationRequest/instrument}}]».
 
     <div class="note">**TODO**: We do not have the RP origin at this step; maybe that should just go in the extension processing steps?</div>
+    Note: This is set in WebAuthn Authenticator, as part of creation? How do we validate/ensure we're not jumping to an incorrect domain? is this currently validated by the browser agent or by the CTAP protocol? Then we need to set it in some way?
+
+    Q: is payeeOrigin optional? What if it's not set? Do we allow it to be empty?
 
 1. Let |extensions| be a new {{AuthenticationExtensionsClientInputs}} dictionary whose
     {{AuthenticationExtensionsClientInputs/payment}} member is set to |payment|, and whose other
@@ -544,6 +591,11 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{PublicKeyCredentialRequestOptions/userVerification}}" → {{UserVerificationRequirement/required}},
     "{{PublicKeyCredentialRequestOptions/extensions}}" → |extensions|]».
 
+  **TODO**: Should the browser agent not extend the 'challenge' object passed in by putting it inside a new JSON structure whihc includes the payment amount and payee details? 
+  
+  **TODO**: Userverification + " required' would probably also come from the WebAuthn value. We cannot have a different auth being 
+  
+  **TODO**: Timeout might have to be a calculation.
 1. For each |id| in `data.credentialIds`:
 
     1. Let |descriptor| be a new {{PublicKeyCredentialDescriptor}} dictionary, initialized as
@@ -559,6 +611,8 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     → |publicKeyOpts|]».
 
     Note: This triggers [[webauthn-3]]'s [[webauthn-3#sctn-getAssertion|Get]] behavior
+    
+    Q: Which version of WebAuthn will SPC require? Can we make it work for Level 1, or is Level 2 or 3 required?
 
 1. Return |outputCredential|.
 
@@ -575,6 +629,8 @@ this extension allows a third-party to perform an authentication ceremony on
 behalf of the [=Relying Party=], and also adds transaction information to the
 signed cryptogram.
 
+Q: Where is the adding of Transaction information defined? I think we should define it in more detail in this spec?
+
 Notably, a website should not call
 {{CredentialsContainer/get()|navigator.credentials.get()}} with this extension
 directly; for authentication the extension can only be accessed via
@@ -586,6 +642,8 @@ directly; for authentication the extension can only be accessed via
 :  Operation applicability
 :: [=registration extension|Registration=] and [=authentication extension|authentication=]
 
+Q: Why does this apply to Registration? Should it not just be for authentication?
+
 :  Client extension input
 :: <xmp class="idl">
     partial dictionary AuthenticationExtensionsClientInputs {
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -189,6 +189,12 @@ example flows and do not limit the scope of how the API can be used.
 This is the first-time flow, in which a new credential is created and stored by
 an issuing bank.
 
+**TODO**: Suggest we add an extra example, which might be more widely used: When a user navigates to bank.com for internet banking and 1) their account access
```
stephenmcgruer, line 192: I'm not opposed to such an example, though from an ideological view I continue to think we shouldn't let historical baggage of complex activations drag down WebAuthn/SPC credential creation :).

There is a slight complication: currently we are unsure whether the WebAuthn WG will allow an SPC credential to be entirely equivalent to a WebAuthn credential (i.e. make them interchangeable), or if they will require an SPC credential to have a special bit so that the RP knows this credential can be used in a third-party auth ceremony.

* If a special bit is required, this example will make a lot of sense. 
* If a special bit is **not** required, this example will just describe normal WebAuthn credential creation and thus be rather redundant imo.

spec.bs
```diff
@@ -212,6 +218,10 @@ an issuing bank.
     credential is created and returned to the issuing bank who stores
     it in their server-side database associated with the user and
     payment instrument(s).
+    
+**TODO**: Define an Agent specific display for Enrollment, indicating that that Payment transactions will be supported (or that this is exclusively for payments)?
```
stephenmcgruer, line 222: Assuming 'Agent specific display' here means 'a browser UX', we (Chrome) oppose this currently. We are happy that the WebAuthn UX that would be shown is sufficient understanding for the user, and we believe that the RP is best suited to build and show any more specific UX that they wish to show regarding payments.

spec.bs
```diff
@@ -212,6 +218,10 @@ an issuing bank.
     credential is created and returned to the issuing bank who stores
     it in their server-side database associated with the user and
     payment instrument(s).
+    
+**TODO**: Define an Agent specific display for Enrollment, indicating that that Payment transactions will be supported (or that this is exclusively for payments)? 
+
+**TODO**: Enhancement to consider: Allow the the user able to select if Payment should be enabled or not (the payment extension) or both?
```
stephenmcgruer, line 224: Again, we (Chrome) currently oppose this.

spec.bs
```diff
@@ -229,6 +239,8 @@ const publicKey = {
   rp: {
     name: "Fancy Bank",
   },
+  
+  Q: Should we perhaps require that  rpId (domain) must be specified if this is inside an iFrame? For top level it's probably OK to use the top level domain?
```
stephenmcgruer, line 243: I'm not immediately sure why one would need to specify rpId here. I believe the WebAuthn spec says to use the callers (iframe's) domain:

https://w3c.github.io/webauthn/#dom-publickeycredentialcreationoptions-rp
```
Its value’s id member specifies the RP ID the credential should be scoped to. If omitted, its value
will be the CredentialsContainer object’s relevant settings object's origin's effective domain. See
§ 5.4.2 Relying Party Parameters for Credential Generation (dictionary PublicKeyCredentialRpEntity)
for further details.
```

Have I overlooked something?

spec.bs
```diff
@@ -255,9 +267,12 @@ const publicKey = {
   authenticatorSelection: {
     userVerification: "required"
   },
+  
+  **TODO**: Extend definition (perhaps in future) to make this optional (so to only have userPresence: "required", and thereby support regions outside of europe) and remove friction of having to type in User PIN's?
```
stephenmcgruer, line 271: I think (though I'd need to double-check) that the spec doesn't currently require any particular `authenticatorSelection` options, though the user agent might (and Chrome currently does).

spec.bs
```diff
@@ -255,9 +267,12 @@ const publicKey = {
...
 
-  timeout: 360000,  // 6 minutes
-
+  timeout: 60000,  // 1 minute
```
stephenmcgruer, line 273: No objection to the change to an example, I just stole this from WebAuthn examples anyway :P

spec.bs
```diff
@@ -282,6 +297,16 @@ navigator.credentials.create({ publicKey })
 
 </div> <!-- non-normative -->
 
+**TODO**: Sugest we provide clarifying text/expand around the following 
+* Does the extension imply that this credential my ONLY be used for Payment, or does this mean the WebAuthn credential could be used for Login or Payment,
```
stephenmcgruer, line 301: Seems reasonable to clarify briefly. It can be used for both login and payment, if the RP wants to. They can also trivially not use it for login; if doing that we recommend they utilize a subdomain (e.g. `payment.bank.com`) to avoid polluting their credential space with credentials they wouldn't accept for login.

spec.bs
```diff
@@ -282,6 +297,16 @@ navigator.credentials.create({ publicKey })
 
 </div> <!-- non-normative -->
 
+**TODO**: Sugest we provide clarifying text/expand around the following 
+* Does the extension imply that this credential my ONLY be used for Payment, or does this mean the WebAuthn credential could be used for Login or Payment, 
+  and both are legitimate?
+* Can the payment extension be 'added' once the webAuthn is created? Or must it be set at registration time? and can it be disabled too?
```
stephenmcgruer, line 303: Currently no, but I think we should think about this. As noted, a lot here depends on the WebAuthn WG's preference/requirements on whether SPC credential == WebAuthn credential, or if a separate bit (of some sort) must be set.

I *think* in general that WebAuthn doesn't allow modifying credentials (maybe I am incorrect), and so its generally a replace operation for something like an 'upgrade'.

spec.bs
```diff
@@ -338,7 +363,9 @@ const request = new PaymentRequest([{
       icon: "https://fancybank.com/card-art.png",
     }, 
 
-    timeout: 360000,  // 6 minutes
+Q: Sugest we define that displayName is Mandatory and icon is optional?
```
stephenmcgruer, line 366: To be clear, you're suggesting a change to https://w3c.github.io/secure-payment-confirmation/#sctn-paymentcredentialinstrument-dictionary ?

I'm not sure offhand whether icon should be optional. Currently all users of SPC wish to pass an icon, and it's much easier to go required --> optional than the other way around, so I would by default lean to keeping both required.

spec.bs
```diff
@@ -358,6 +387,8 @@ try {
 
   // response.data is a PublicKeyCredential, with a clientDataJSON that
   // contains the transaction data for verification by the issuing bank.
+  
+  Q: We need to define the format of Response.data somewhere in this doc, including what fields are present if the user accepts, of if they cancel/timeout.
```
stephenmcgruer, line 391: https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-respond-to-a-payment-request

spec.bs
```diff
@@ -391,6 +422,8 @@ below and in [[#index-defined-elsewhere]].
     away from its Payment Request origins. For developers, this should
     improve feature detection, invocation, and other aspects of the
     API.</div>
+    
+    Q: Is the suggestion to create SPC specific methods, or roll everything into WebAuthn, with no custom-UI/logic in the browser agent?
```
stephenmcgruer, line 426: I'll leave the question of API shape for now, but note that we (Chrome) currently expect to always present payment-specific UX for an SPC authentication, no matter the API shape to call into it.

My personal understanding is that this NOTE only seeks to change the JavaScript code used to call SPC but none of the current behavior or UX would change.

spec.bs
```diff
@@ -464,18 +499,27 @@ members:
...
     :: The number of milliseconds before the request to sign the transaction details times out. At most 1 hour.
+    
+    **TODO**: We need to define how this timeout is calculated. E.g. If the timeout is 60 seconds, and the uer spends 25 seconds on the SPC payment page in the browser, does the WebAuthn request then get initialized with 35 seconds or with 60 seconds? 
+    
+    **TODO**: What do we do if the SPC page is open for more than 60 seconds? Do we cancel without attempting WebAuthn?
 
     :  <dfn>payeeOrigin</dfn> member
     :: The origin of the payee that this SPC call is for (e.g., the merchant).
+    
+    **TODO**: Will this be validated in any way?
```
stephenmcgruer, line 521: It is impossible for the UA to validate this in the redirect case. This is identical to displayName and icon, where we are just relaying information from the SPC caller to the user, and then signing it to tell the RP that this is what we showed to the user. It's up to the RP to decide if it trusts the SPC caller to be doing a transaction for the payeeOrigin.

spec.bs
```diff
@@ -464,18 +499,27 @@ members:
...
+    
+    **TODO**: We need to define how this timeout is calculated. E.g. If the timeout is 60 seconds, and the uer spends 25 seconds on the SPC payment page in the browser, does the WebAuthn request then get initialized with 35 seconds or with 60 seconds? 
+    
+    **TODO**: What do we do if the SPC page is open for more than 60 seconds? Do we cancel without attempting WebAuthn?
 
     :  <dfn>payeeOrigin</dfn> member
     :: The origin of the payee that this SPC call is for (e.g., the merchant).
+    
+    **TODO**: Will this be validated in any way? 
+    **TODO**: Will this be inserted by the browser agent if not provided?
```
stephenmcgruer, line 522: The conclusion in https://github.com/w3c/secure-payment-confirmation/pull/89 was to make it required, though I see you had a comment in that thread that I may have misunderstood (I thought you were saying the `topOrigin` field from your first bullet-point should be auto-filled, but perhaps you meant `payeeOrigin` should be auto-filled)

spec.bs
```diff
@@ -533,6 +577,9 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{AuthenticationExtensionsPaymentInputs/instrument}}" → |data|.{{SecurePaymentConfirmationRequest/instrument}}]».
 
     <div class="note">**TODO**: We do not have the RP origin at this step; maybe that should just go in the extension processing steps?</div>
+    Note: This is set in WebAuthn Authenticator, as part of creation? How do we validate/ensure we're not jumping to an incorrect domain? is this currently validated by the browser agent or by the CTAP protocol? Then we need to set it in some way?
+
+    Q: is payeeOrigin optional? What if it's not set? Do we allow it to be empty?
```
stephenmcgruer, line 582: It is not optional; https://w3c.github.io/secure-payment-confirmation/#dictdef-securepaymentconfirmationrequest

spec.bs
```diff
@@ -533,6 +577,9 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{AuthenticationExtensionsPaymentInputs/instrument}}" → |data|.{{SecurePaymentConfirmationRequest/instrument}}]».
 
     <div class="note">**TODO**: We do not have the RP origin at this step; maybe that should just go in the extension processing steps?</div>
+    Note: This is set in WebAuthn Authenticator, as part of creation? How do we validate/ensure we're not jumping to an incorrect domain? is this currently validated by the browser agent or by the CTAP protocol? Then we need to set it in some way?
```
stephenmcgruer, line 580: Can you rephrase this? I'm not sure what you're trying to ask here.

spec.bs
```diff
@@ -544,6 +591,11 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{PublicKeyCredentialRequestOptions/userVerification}}" → {{UserVerificationRequirement/required}},
     "{{PublicKeyCredentialRequestOptions/extensions}}" → |extensions|]».
 
+  **TODO**: Should the browser agent not extend the 'challenge' object passed in by putting it inside a new JSON structure whihc includes the payment amount and payee details?
```
stephenmcgruer, line 594: No, that is the purpose of the extension to the CollectedClientData, to explicitly avoid hacking the `challenge`: https://w3c.github.io/secure-payment-confirmation/#sctn-collectedclientpaymentdata-dictionary

spec.bs
```diff
@@ -544,6 +591,11 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     "{{PublicKeyCredentialRequestOptions/userVerification}}" → {{UserVerificationRequirement/required}},
     "{{PublicKeyCredentialRequestOptions/extensions}}" → |extensions|]».
 
+  **TODO**: Should the browser agent not extend the 'challenge' object passed in by putting it inside a new JSON structure whihc includes the payment amount and payee details? 
+  
+  **TODO**: Userverification + " required' would probably also come from the WebAuthn value. We cannot have a different auth being
```
stephenmcgruer, line 596: Cut-off sentence?

spec.bs
```diff
@@ -559,6 +611,8 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     → |publicKeyOpts|]».
 
     Note: This triggers [[webauthn-3]]'s [[webauthn-3#sctn-getAssertion|Get]] behavior
+    
+    Q: Which version of WebAuthn will SPC require? Can we make it work for Level 1, or is Level 2 or 3 required?
```
stephenmcgruer, line 615: I think L2 might be required for discoverable credentials?

spec.bs
```diff
@@ -575,6 +629,8 @@ this extension allows a third-party to perform an authentication ceremony on
 behalf of the [=Relying Party=], and also adds transaction information to the
 signed cryptogram.
 
+Q: Where is the adding of Transaction information defined? I think we should define it in more detail in this spec?
```
stephenmcgruer, line 632: https://w3c.github.io/secure-payment-confirmation/#sctn-payment-extension-enrollment

Under 'Client extension processing (authentication)', where it modifies step 9.

spec.bs
```diff
@@ -586,6 +642,8 @@ directly; for authentication the extension can only be accessed via
 :  Operation applicability
 :: [=registration extension|Registration=] and [=authentication extension|authentication=]
 
+Q: Why does this apply to Registration? Should it not just be for authentication?
```
stephenmcgruer, line 645: As per 'Client extension processing (registration)' below, it needs to relax the ban on creation in a cross-origin iframe.

spec.bs
```diff
@@ -595,11 +653,14 @@ directly; for authentication the extension can only be accessed via
     dictionary AuthenticationExtensionsPaymentInputs {
       boolean isPayment;
 
+Q: When would this be set to False? Is this not true by the virtue of being present?
```
stephenmcgruer, line 656: Yes, this is just an awkward outcome of WebIDL treating an empty dict as false (afaik). So if during registration you did:

```
const publicKeyCreationOptions = {
  ...
  extensions: {
    payment: {},
  },
};
navigator.credentials.create({publicKey: publicKeyCreationOptions});
```

Then afaik the payment extension is actually *not* set. So you need some way to indicate 'no I do want this'.

As per the note under https://w3c.github.io/secure-payment-confirmation/#sctn-enrollment, if WebAuthn WG allow credential creation in a 3p context, and if we get to a point where the browser doesn't have to cache the credentials at creation time, this extension will no longer need to be registration time too, and this field can be dropped.

(`isPayment` definitely needs  better documentation. It is mentioned here https://w3c.github.io/secure-payment-confirmation/#dom-authenticationextensionspaymentinputs-ispayment but not well)

spec.bs
```diff
@@ -642,6 +703,8 @@ directly; for authentication the extension can only be accessed via
          (see [[#sctn-permissions-policy]]). We could additionally require and
          consume a [=transient activation=] here, if we felt the permission policy
          is not sufficient.
+         
+         Note: I feel we must still check that the ID's corresponds to the domain that was provided; you cannot use a key with Bank2 if Bank1 claims to be asking for it? The RP must be included in the signature at least.
```
stephenmcgruer, line 707: The RP is in the signature; https://w3c.github.io/secure-payment-confirmation/#dom-collectedclientadditionalpaymentdata-rp

spec.bs
```diff
@@ -656,6 +719,8 @@ directly; for authentication the extension can only be accessed via
         1. Skip step 6.1, which compares *options.rpId* to *effectiveDomain*
 
             Note: This enables cross-domain authentication ceremonies; see [[#sctn-use-case-merchant-authentication]].
+            
+            Q: Should we not at least compare to the claimed Bank Identity? How else will we know the value displayed is actually linked to the credential?
```
stephenmcgruer, line 723: Because currently the caller of SPC doesn't set the RP origin shown to the user. There is no input to https://w3c.github.io/secure-payment-confirmation/#sctn-securepaymentconfirmationrequest-dictionary where one can 'claim' the RP value.

How the user agent actually *gets* the RP's identity from the credential is an outstanding issue for the spec. In the Chrome impl, we cache it at creation time. In the future, I hope to retrieve it via the Conditional UI concept in WebAuthn (which will also require the ability to fetch an RP for a given credential).

We *could* ask the caller to make an assertion about what domain it is, but I think given that there are technical solutions to avoid that, we should. If we do ultimately ask the caller to make such a claim, then yes we should verify that once we know the value.

spec.bs
```diff
@@ -724,6 +791,8 @@ fields:
     :: The instrument information that was displayed to the user.
 </dl>
 
+Q: We should probably state that these fields must be displayed to the user.
```
stephenmcgruer, line 794: See the big TODO in https://w3c.github.io/secure-payment-confirmation/#sctn-transaction-confirmation-ux ;)

spec.bs
Goosth: If we're aligning to Webauthn, then I would suggest we remove the detailed explanation and flows. It would be better to rather just refer to the WebAuthn document and process for Enrollment.

We would then just
* give high level explanation on the use-cases (the two mentioned)
* Re-iterate that this is standard WebAuthn and there is no cross domain capability in registration, so the standard WebAuthn process will be followed with no UI.

spec.bs
Goosth: Ok. Let's then just state that (see my previous comment).

spec.bs
Goosth: Given that this is a Registration and there is no cross domain and no custom display, I agree that this should be fine.

spec.bs
Goosth: That's my understanding as well. Let's state this is the situation right now. (Do you think we need to verify this with WebAuthn team?)

spec.bs
Goosth: Thanks.

spec.bs
Goosth: So would Chrome then be deemed as compliant if it only supports a subset?
Is it your intention to support the other options as well in future; just not initially?

spec.bs
Goosth: I guess both should be filled. But agree on the comment that it's not provable. Perhaps better to, similar to icon and displayName and make it mandatory.

spec.bs
Goosth: Saw it. :-)

spec.bs
Goosth: Will you be able tp make it work without that or is discoverable credentials a minimum requirement?

spec.bs
ianbjacobs: Hi @goosth, I think there is a cross-origin iframe registration capability in SPC, without special UI in the current Chrome implementation.

spec.bs
stephenmcgruer: I will be speaking at the WebAuthn WG tomorrow afternoon EST, and I expect a resolution to come out of that.

spec.bs
stephenmcgruer: > So would Chrome then be deemed as compliant if it only supports a subset?

Nope! :D

> Is it your intention to support the other options as well in future; just not initially?

I think this will go one of two ways: either we will hear no demand for other options and may restrict SPC for simplicity and/or clarity, or we will hear a compelling use-case and Chrome will add support for other options.

spec.bs
stephenmcgruer: I would consider them a minimum requirement, in order to move from a browser-based cache to a platform-based cache and thus avoid SPC credentials being per-browser on a given device. See https://github.com/w3c/secure-payment-confirmation/issues/92

spec.bs
Goosth: Agreed. But that is a WebAuthn standard. You still need the iframe to be opened in the RP's domain and the iframe to be given the correct permissions by the merchant. 
It's not using the SPC superpower of being called by the merchant and creating a key in the RP's domain.

spec.bs
Goosth: I guess in that case then we'll need to pair back the spec at a later stage
@ianbjacobs : What is the best advice 
1.  have the spec adhere more to Chrome's initial implementation and note this as a future extension
2. have the spec be more generic and not have Chrome fully implement it?

spec.bs
Goosth: If the claimed RP is different than the RP that created it, why would the Browser/WebAuthn not also verify it?
First check that the RP is correct before asking the Authenticator to sign it?

stephenmcgruer: Sorry, I should have quoted to indicate the part I was replying to: "The RP must be included in the signature at least."

The caller of SPC makes no claim on the RP (it's not [an input to the API](https://w3c.github.io/secure-payment-confirmation/#sctn-securepaymentconfirmationrequest-dictionary)). They pass in a list of credential IDs, which do not contain RP information. The browser/authenticator* then matches the credential IDs to the underlying RP, which is what is signed.

(One thing the spec doesn't deal with today which we should fix - what if you pass in credential IDs from multiple RPs, what happens. This is a silly edge-case that doesn't really matter, but we should spec it to just fail in a privacy-preserving way probably).

* Today via the browser-stored metadata from credential creation time, in the long-term this should be queryable from the authenticator.

ianbjacobs:
@Goosth, I haven't checked in lately to see whether we've addressed issues here to your satisfaction.

ianbjacobs:
Closing this issue due to inactivity and the fact that many points have been taken into account separately.

PR #105, opened 2021-08-16
[Spec] Require that `payeeOrigin` is fully qualified

stephenmcgruer:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/105.html" title="Last updated on Aug 17, 2021, 6:54 PM UTC (90a3c28)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/105/6173bf7...90a3c28.html" title="Last updated on Aug 17, 2021, 6:54 PM UTC (90a3c28)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -328,7 +328,9 @@ const request = new PaymentRequest([{
     instrument: {
       displayName: "Fancy Card ****1234",
       icon: "https://fancybank.com/card-art.png",
-    }, 
+    },
+
+    payeeOrigin: "merchant.com",
 
     timeout: 360000,  // 6 minutes
   }], {
--- a/spec.bs
+++ b/spec.bs
@@ -54,6 +54,11 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: client extension; url: client-extension
         text: registration extension; url: registration-extension
         text: authentication extension; url: authentication-extension
+
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
+    type: dfn
+        urlPrefix: origin.html
+            text: origin; url: concept-origin
 </pre>
 
 <div class="non-normative">
@@ -330,7 +335,7 @@ const request = new PaymentRequest([{
       icon: "https://fancybank.com/card-art.png",
     },
 
-    payeeOrigin: "merchant.com",
+    payeeOrigin: "https://merchant.com",
 
     timeout: 360000,  // 6 minutes
   }], {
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -336,7 +336,9 @@ const request = new PaymentRequest([{
     instrument: {
       displayName: "Fancy Card ****1234",
       icon: "https://fancybank.com/card-art.png",
-    }, 
+    },
+
+    payeeOrigin: "merchant.com",
```
rsolomakhin, line 341: Quick question from the stand point of an implementer: Do we require the payee origin to be https://, or just domain will be enough? If latter, then we would be assuming https:// for the scheme, right? The scheme (provided or assumed) would be useful for origin validation.

stephenmcgruer: Great question, actually. I wasn't thinking properly about the fact that an origin is more than its domain... (it's potentially the scheme, port, host, and domain).

`payeeOrigin` is not a verified concept (by the browser), so I'm not sure adding any requirements on the value here makes sense? E.g. `psp.com` could claim payeeOrigin to be `https://merchant.com` or `http://merchant.com`, and you as the RP don't actually know if **either** of those were involved in the payment, just that the browser showed the user `merchant.com` (is that what we show? or do we show full domain?).

rsolomakhin: We show only "merchant.com" because that's the current recommendation for the browser UX security team. That may change in the future.

Even though "merchant.com" is used only for display and in `"clientDataJson"`, the security team also would want us to keep the origin in a `url::Origin` type, so it's clear that it's not an arbitrary string and it can be validated. Constructing `url::Origin` from a domain typically requires a scheme, so that's why I'm asking this question.

I'm not as concerned about the port number, because a scheme typically has a default port.

rsolomakhin: Anywho, for simplicity of the spec and implementation, it may be preferable to require `"payeeOrigin"` to always have the `"https://"` scheme, at least for now.

stephenmcgruer: Makes sense to me; updated the PR.

I also noticed that the spec has [rpOrigin](https://w3c.github.io/secure-payment-confirmation/#dom-authenticationextensionspaymentinputs-rporigin) which I realized should probably be just `rp` or `rpId`? (In either case being the *id* of the RP, not the origin)

rsolomakhin: The RP field is called `rp` in the current implementation and it is indeed the _id_, not the origin, because it is inherited from `navigator.credentials.create({publicKey})`.

rsolomakhin (approved):

PR #112, opened 2021-08-23
[Explainer] Re-write the explainer to best-practices, part 2/n

stephenmcgruer:
This series of commits both brings the explainer up-to-date with the current spec,
and re-writes it to follow the guidance at https://w3ctag.github.io/explainers

Part 2 adds the Security and Privacy considerations.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -398,7 +398,7 @@ ensure that the use of the assertion matches the expected interaction type.
 
 In this attack, a malicious third-party tries to use SPC (again with either
 legimitately or illictly obtained credentials) to initiate an unauthorized
-payment. The attack is infeasible:
+payment. Such an attack has a low chance of success for several reasons:
 
 * When the attacker initiates SPC, the user will be shown an interface by the
     User Agent that clearly states the transaction details (including the payee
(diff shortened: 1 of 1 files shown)


explainer.md
```diff
@@ -362,21 +363,147 @@ merchants that accept credit cards.
...
+
+In addition to the above, a Relying Party should also record what type of
+interaction (i.e. login or payment) a given `challenge` is generated for and
+ensure that the use of the assertion matches the expected interaction type.
+
+**Payment attack**
+
+In this attack, a malicious third-party tries to use SPC (again with either
+legimitately or illictly obtained credentials) to initiate an unauthorized
+payment. The attack is infeasible:
```
ianbjacobs, line 401: Editorial: Change "The attack is infeasiable" to "Such an attack has a low chance of success for several reasons:"

rsolomakhin (approved):

explainer.md
```diff
@@ -362,21 +363,147 @@ merchants that accept credit cards.
...
+For example, the Merchant could tell the Account Provider (in the backend) that
+it is initiating a purchase of $100, but then pass $1 to the SPC API (and thus
+show the Customer a $1 transaction to verify).
+
+This form of attack already exists on the web today (where an Account Provider
+largely has to trust that the Merchant is showing the correct details), and SPC
+actually makes it harder to perform this attack. After receiving an assertion,
+the Account Provider can (and should) verify the payment data in the assertion
+to ensure it aligns with the transaction details provided to the Account
+Provider by the Merchant.
```
ianbjacobs, line 437: The text as written does not explain how the Account Provider previously received information about the transaction from the Merchant. My understanding is that this is implementation dependent, e.g., when integrating SPC with 3DS, transaction info is provided early to the ACS, then validated later by the ACS. Perhaps something like this:

This form of attack already exists on the web today (where an Account Provider largely has to trust that the Merchant is showing the correct details).  We anticipate that SPC will provide better protection against this attack when used according to this common pattern:

* The Merchant provides some "pre-authentication" transaction information (including merchant name and amount) to the Account Provider over backend rails.
* After receiving an assertion, the Account Provider can (and should) compare the pre-authentication details with the assertion details to ensure they align.

ianbjacobs (approved):
See my small suggestions. Thanks!

PR #114, opened 2021-08-23
[Spec] Write the 'transaction UX' section

stephenmcgruer:
The goal here is to enforce that User Agents do communicate the necessary
information that will be signed over, but not to restrict their implementation
of how to do so.


stephenmcgruer:
@domfarolino - I've straw-personed some minimal text here as a starting point. I'd love your thoughts on the best way to specify this without constraining UAs.

ianbjacobs (approved):

PR #116, opened 2021-08-24
[TAG] Update the security/privacy questionnaire

stephenmcgruer:


security-privacy-questionnaire.md
```diff
@@ -88,13 +86,16 @@ created by [Web Authentication].
...
+This feature mostly inherits the behavior from [Web Authentication], with the
+following exceptions:
 
-- A `PaymentCredential` can only be created in a first-party secure context.
-- A `PaymentCredential` can be exercised in any context where Payment Request
-  API is allowed, i.e. a first-party secure context, or a third-party secure
-  context with [delegated permission] from the top-level context.
+- A SPC-enabled `PublicKeyCredentials` can be created in a third-party secure
+  context, **for that third-party**, if given [delegated permission] from
+  the top-level context.
```
rsolomakhin, line 94: I know that this an accurate description of a secure and privacy preserving action, but perhaps this can rephrased to mention the "RP" keyword? For example:

> When an RP is embedded in a cross origin iframe with the [delegated permission], this RP can register an SPC-enabled `PublicKeyCredential` for itself. 

Or something like that. Wordsmithing is hard.

rsolomakhin (approved):
LGTM % nits

security-privacy-questionnaire.md
```diff
@@ -3,20 +3,14 @@
...
-Minimizing redirection or any form of friction is critical to avoid cart
-abandonment during a checkout process. Direct authentication using the issuing
-bank's credential is necessary because the merchant or PSP may be fradulent.
-
-This feature also allows a web site to find out that a given credential does not
-exist - the `paymentRequest.show()` call will reject if no matching credential
-is found. We believe this does not pose a significant probing risk because
-`paymentRequest.show()` consumes a user activation. This makes it hard for a
-malicious merchant to automate probing at scale.
+`PublicKeyCredential` to exercise the credential via the Payment Request API to
```
ianbjacobs, line 6: s/via the Payment Request API/in order

(No need to mention PR API here and also we hope that changes!)

security-privacy-questionnaire.md
```diff
@@ -3,20 +3,14 @@
...
+sign over transaction data. This is to allow a merchant or a Payment Service
+Provider (PSP) that represents the merchant to request strong authentication of
+the payment instrument used in the transaction without redirecting to the
+issuing bank of the payment instrument. Minimizing redirection or any form of
+friction is critical to avoid cart abandonment during a checkout process.
+Direct authentication using the issuing bank's credential is necessary because
+the merchant or PSP may be fradulent.
```
ianbjacobs, line 13: Suggest changing "Direct authentication using the issuing bank's credential is necessary because the merchant or PSP may be fradulent." to:

"In many payment systems, the Account Provider wishes (or is required by regulation) to authenticate the user for fraud protection; SPC allows them to fulfill the requirement to validate the authentication while enabling the merchant to manage the user experience."

security-privacy-questionnaire.md
```diff
@@ -28,26 +22,22 @@ This feature does not collect or expose any such information.
...
 
-The only sensitive information this feature handles is an opaque string that can
-be mapped to a payment instrument by an issuing bank. This instrument ID is
-saved in the user agent along side the public key credential when the issuing
-bank creates a `PaymentCredential`. Later, an origin that has access to this
-instrument ID, presumably via a trusted server integration with the issuing
-bank, can provide it to the user agent via [Payment Request API] to exercise the
-corresponding `PaymentCredential`.
+The only sensitive information this feature handles is the public ID of the
+created credential. This credential ID is saved in the user agent along side
```
ianbjacobs, line 26: Suggest:

"The only sensitive information this feature handles relates to the (WebAuthn) credential ID of the created credential."

security-privacy-questionnaire.md
```diff
@@ -28,26 +22,22 @@ This feature does not collect or expose any such information.
...
+it to the user agent via [Payment Request API] to exercise the corresponding
+`PublicKeyCredential`.
 
 ## 5. Does this specification introduce new state for an origin that persists across browsing sessions?
 
-This feature allows an issuing bank to persist an icon, a descriptive string and
-an issuer-defined payment instrument ID with a public key credential when it
-creates a new `PaymentCredential`. However, this data is only used by the
-user agent to render UI to the user, and cannot be queried via JavaScript.
+No.
```
ianbjacobs, line 35: Hmm, isn't the answer "Yes, but (1) we expect that to go away with discoverable credentials, and (2) the information is only available through the API." ?

security-privacy-questionnaire.md
```diff
@@ -76,10 +66,18 @@ No.
...
+
+* Payment instrument icon and string
+* Payment amount and currency
+* Intended payee
+
+This information is displayed to the user in the user agent's native UI when
+this credential is exercised via Payment Request API. It could be spoofed by
+the merchant, however the issuing bank that receives the assertion will be able
+to detect this and reject the payment (and take action against the merchant as
+appropriate).
```
ianbjacobs, line 80: Do you want to also say: 

"It could be in part spoofed by the merchant, but since the UX is browser controlled, it is rendered atop the chrome (which cannot be done via JavaScript)" 

(or however you would say that.)

It might be worth mentioning even if it's not the primary mitigation observation.

ianbjacobs (approved):
Hi @stephenmcgruer,

If my suggestions work for you, please consider this approved. Thank you!

security-privacy-questionnaire.md
stephenmcgruer: What state is that? A WebAuthn credential already persists across browsing sessions.

security-privacy-questionnaire.md
stephenmcgruer: That is not relevant to the question, which is about site control over browser UX. The idea of a site spoofing browser UX (which does not involve controlling it!) is true for every browser UX in existence (we see it for notifications, for WebAuthn, for ...)

ianbjacobs: So what do you mean when you say "spoofed by the merchant"?

stephenmcgruer: The merchant can input incorrect data into the SPC API. Possibly spoofing is the wrong word there.

security-privacy-questionnaire.md
ianbjacobs: Point taken. (I admit to not knowing the details about WebAuthn implementations.)

security-privacy-questionnaire.md
ianbjacobs: How about: "If the merchant provides incorrect or fraudulent data to the API for confirmation by the user, the Relying Party can detect this when validating the assertion and reject the transaction."

stephenmcgruer: SGTM, done.

PR #155, opened 2021-11-02
Describe protections against "wrong credential" attack

ianbjacobs:
New text to cover the attack where a caller intentionally or accidentally calls the API
with an instrument string and completely unrelated (but valid) SPC credentials.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/155.html" title="Last updated on Nov 2, 2021, 9:52 PM UTC (3d4204e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/155/4015745...3d4204e.html" title="Last updated on Nov 2, 2021, 9:52 PM UTC (3d4204e)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1095,6 +1095,11 @@ an unauthorized payment using Secure Payment Confirmation credentials
 * If the [=Relying Party=] is expecting a transaction, it will detect an
     unfamiliar `challenge` and reject the assertion.
 
+The same backend protections apply in the scenario where the caller
+intentionally or mistakenly provides the API with instrument
+information and Secure Payment Confirmation credentials not related
+through registration.
+
 ## Merchant-supplied authentication data ## {#sctn-security-merchant-data}
 
 A consequence of this specification's third-party authentication ceremony is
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
Isn't this just 10.2? (https://w3c.github.io/secure-payment-confirmation/#sctn-security-merchant-data)

ianbjacobs:
@stephenmcgruer,

It's sort of 10.1 and sort of 10.2. We haven't called out the case of "wrong credentials" yet. If you think this fits better in 10.2 I'm happy to move it there.

stephenmcgruer:
Hrm, it sort of depends which way you look at it 😂 . I view this as a spoofing attack where you're spoofing the instrument descriptor, whereas you're describing it as a ??? attack where from your view the instrument is valid but the credentials are wrong.

Probably fits in 10.2, as a paragraph after "This could lead to a spoofing attack", saying something like 'the caller may also use the correct details for the transaction but pass a credential that doesn't match what the RP expects'.

There's an undertone of assuming that a credential is for a given payment instrument (or set of instruments) in this PR, and I just want to re-mention that SPC makes no strong statement about payment instruments - to SPC, payment instrument details is just another blob of data as interesting (or not) as e.g. the payeeOrigin. It's entirely up to the RP to decide what they accept for a given credential.

ianbjacobs:
@stephenmcgruer, let me know if this fix suits you. Thanks!

stephenmcgruer (approved):

Issue #127, opened 2021-09-07
Accessible icon information


ianbjacobs:
For users with some visual disabilities, the icon may not be accessible. Would it make sense to support an "alt" string as input (e.g., that the browser can make available in the UX or to assistive technologies)?

rsolomakhin:
I would consider the instrument display name to be this alt text.

ianbjacobs:
The instrument string does not convey the same information. A card number does not tell you the issuing institution (for example).

rsolomakhin:
An alt text for the screen reader could make sense if the icon's intention is to convey entirely different information from its display name.

ianbjacobs:
@RealJoshue108,

My expectation is that the icon will tell the user about the issuer of the payment instrument, for example "Mastercard" or "Barclays". Thus, the alt text might be read as "Mastercard ***1234".

stephenmcgruer:
I would love to hear from e.g. @jcemer-stripe , @davordavidovic, or other users of SPC on whether having alt-text for the icon would provide **additional** explanation value on top of the displayName, or whether its just visual 'eye-candy'.

For example, if most displayNames are `NetworkName ****1234` anyway, and the alt text for the icon would also be e.g. `NetworkName icon`, then to me not much value is added. On the other hand, if displayNames are often something like `Stephen's card ****1234` then the icon is distinct and an alt-text makes a lot of sense.

jcemer-stripe:
In the context of card payment, the icon is used more as a 'eye-candy'. I don't see much value of adding an alt-text for cards but it could be a nice touch for other use cases.

rsolomakhin:
> eye-candy

Do you mean to say that it's not important for accessibility / for understanding the context of the payment?

jcemer-stripe:
> > eye-candy
> 
> Do you mean to say that it's not important for accessibility / for understanding the context of the payment?

I cannot make that strong statement but I think that for card payments the icon is pretty much the card brand which is in general already present in the display name.

ianbjacobs:
Are there use cases where one might do something like this:

[Bank icon] Mastercard *****1234

That seems to be the main use case.

ianbjacobs:
@RealJoshue108, please see pull request #162, which recommends using the displayName to ensure the accessibility of the icon (rather than adding another alt value). According to comments above, the displayName sounds like it will suffice (or can be made to do so). thanks!

michael-n-cooper:
[...]

PR #156, opened 2021-11-05
Alt for icon

ianbjacobs:
To improve accessibility and address #127, a proposal add an optional "alt" to instrument details when displayName is not sufficient to describe the icon.

P.S. I named the pull request 217 instead of 127. Sorry!

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/156.html" title="Last updated on Nov 8, 2021, 1:59 PM UTC (851cc4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/156/cd8781c...851cc4c.html" title="Last updated on Nov 8, 2021, 1:59 PM UTC (851cc4c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -872,14 +872,8 @@ contains the following members:
          of the {{CollectedClientAdditionalPaymentData}} structure.
 
     :  <dfn>alt</dfn> member
-    :: An optional member that provides a text description of the icon.
-       This member should be used when
-       {{PaymentCredentialInstrument/displayName}} does not already
-       include a sufficient text description of the icon. For example,
-       if {{PaymentCredentialInstrument/displayName}} includes a card
-       payment network name but the icon is that of the issuing bank,
-       use this member to provide a text version of the name of the
-       bank.
+    :: An optional member that provides a text description of the icon. This member should be used when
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -842,6 +842,7 @@ The following data structures are shared between registration and authentication
     dictionary PaymentCredentialInstrument {
         required DOMString displayName;
         required USVString icon;
+                 USVString alt;
```
stephenmcgruer, line 845: This should probably be a `DOMString` unless you have a good reason for using a `USVString`.

ianbjacobs:
@r12a and @aphillips, please indicate whether dir/lang are required for an "alt" attribute. 

Also, would it suffice to have dir/lang at the level of the dictionary that would apply to both displayName and alt?

r12a:
I have certainly seen instances in the past where lang and/or dir were needed (but unavailable) in HTML for an alt (or title) because the alt text in a different language or had direction sensitive text that differed from the element text.  We were unable to convince the HTML5 folks to change the HTML approach because they just wanted to pave the cowpaths (and so  the ITS spec explains a workaround for dealing with such an eventuality in HTML/XML).  That was unfortunate, because we had not long before convinced the XHTML2 folks to use an element instead of an attribute for user-readable types of text (including alt).

> Also, would it suffice to have dir/lang at the level of the dictionary that would apply to both displayName and alt?

So, no.  That won't cut it for all cases.

ianbjacobs:
Closing this one in favor of #162.

PR #167, opened 2021-12-06
[Spec] Better define what an SPC credential is and its requirements

stephenmcgruer:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/167.html" title="Last updated on Dec 6, 2021, 7:17 PM UTC (b3ff98c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/167/c5f5828...b3ff98c.html" title="Last updated on Dec 6, 2021, 7:17 PM UTC (b3ff98c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -398,32 +398,33 @@ try {
 # Terminology # {#sctn-terminology}
 
 : <dfn export>SPC Credential</dfn>
-:: A WebAuthn credential that can be used for the behaviors defined in this specification.
-    Can be used as a normal credential (e.g. for authentication flows), as well
-    as for SPC behavior. A [=Relying Party=] must explicitly opt-in to making an
-    SPC credential during credential creation.
+:: A [=public key credential|WebAuthn credential=] that can be used for the
+    behaviors defined in this specification. When an SPC Credential is to be
+    used by a party other than the [=Relying Party=], the [=Relying Party=] must
+    explicitly opt in by declaring the credential for SPC at creation time.
+
+    This specification does not intend to limit how SPC credentials may (or may
+    not) be used by a [=Relying Party=] for other authentication flows (e.g.,
+    login).
 
 : <dfn>Steps to determine if a credential is SPC-enabled</dfn>
 :: An as-yet undefined process by which a user agent can, given a [=credential
     ID=], determine if the credential represented by that ID is an [=SPC
     Credential=].
 
-    This allows the user agent to ensure that the powers of an SPC credential 
-    are only applied to credentials created as SPC-enabled.
-
     NOTE: See <a href="https://github.com/w3c/webauthn/issues/1667">WebAuthn
     issue 1667</a>.
 
-: <dfn>Steps to silently determine if a credential is valid for the current device</dfn>
+: <dfn>Steps to silently determine if a credential is available for the current device</dfn>
 :: An as-yet undefined process by which a user agent can, given a [=credential
     ID=], silently (i.e., without user interaction) determine if the credential
-    represented by that credential ID is valid for the current device (i.e.,
+    represented by that credential ID is available for the current device (i.e.,
     could be successfully used as part of a WebAuthn
     [[webauthn-3#sctn-getAssertion|Get]] call).
 
-    This allows the user agent to only conditionally display a transaction UX to
-    the user if there is some chance that they can successfully complete the
-    transaction.
+    This allows the user agent to only conditionally display
+    [[#sctn-transaction-confirmation-ux|the transaction UX]] to the user if
+    there is some chance that they can successfully complete the transaction.
 
     NOTE: This property will likely require that [=SPC Credentials=] be
     [=discoverable credential|discoverable=]; as such this specification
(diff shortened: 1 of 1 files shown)


rsolomakhin (approved):

stephenmcgruer:
@ianbjacobs - I would also be interested in your review here as someone less tied into the technical details of SPC. I'm trying to make it clearer what our needs are for SPC credentials, and roughly how that ties into how SPC works.

spec.bs
```diff
@@ -68,6 +70,15 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
         text: WebDriver error code; url: dfn-error-code
 </pre>
 
+<pre class="biblio">
+{
+    "webauthn-conditional-ui": {
+        "title": "WebAuthn Conditional UI Proposal",
+        "href": "http://www.example.com/foo-bar/"
```
ianbjacobs, line 77: Do you mean this URI?
 https://github.com/w3c/webauthn/pull/1576

stephenmcgruer: Hah, good catch! I said to myself I'll fix that later... and never did. Fixed now.

spec.bs
```diff
@@ -384,30 +395,43 @@ try {
 }
 </pre>
 
-# Dependencies # {#sctn-dependencies}
+# Terminology # {#sctn-terminology}
+
+: <dfn export>SPC Credential</dfn>
+:: A WebAuthn credential that can be used for behavior defined in this spec.
```
ianbjacobs, line 401:
```suggestion
:: A WebAuthn credential that can be used for the behaviors defined in this specification.
```

spec.bs
```diff
@@ -384,30 +395,43 @@ try {
...
+    Can be used as a normal credential (e.g. for authentication flows), as well
+    as for SPC behavior. A [=Relying Party=] must explicitly opt-in to making an
+    SPC credential during credential creation.
+
```
ianbjacobs, line 405: I'm not sure to agree fully with the opt-in characterization. 

We've heard a desire by some (not all) for an RP to be able to use a vanilla credential for SPC when in a 1p context. In that case, I would argue that they are opting in by using the credential at authentication time. I agree that for cross-origin usage we've settled on creation-time opt-in.

I also think that this specification should not say that an SPC can be used for other authentication flows. My understanding is that we are still discussing that. At the current time I would prefer to say that this spec does not intend to constrain how SPC credentials may or may not be used in other contexts.

With all that in mind, here's a proposed change:

"When an SPC Credential is to be used by a party other than the [=Relying Party=], the [=Relying Party=] must explicitly opt in by declaring the credential for SPC at creation time. This specification does not intend to limit how SPC credentials may (or may not) be used by a [=Relying Party=] for other authentication flows (e.g., login)."

In that proposal I remain silent on RP's using vanilla WebAuthentication credentials as SPC credentials in a 1p context. Once we resolve that issue more completely we may wish to speak to it explicitly.

spec.bs
```diff
@@ -384,30 +395,43 @@ try {
...
+: <dfn>Steps to determine if a credential is SPC-enabled</dfn>
+:: An as-yet undefined process by which a user agent can, given a [=credential
+    ID=], determine if the credential represented by that ID is an [=SPC
+    Credential=].
 
-This specification relies on several other underlying specifications, listed
-below and in [[#index-defined-elsewhere]].
+    This allows the user agent to ensure that the powers of an SPC credential 
+    are only applied to credentials created as SPC-enabled.
```
ianbjacobs, line 412: I think you can delete:

"This allows the user agent to ensure that the powers of an SPC credential are only applied to credentials created as SPC-enabled."

I think the intention is carried by the preceding sentence, and that the preceding sentence allows for the case of WebAuthn-in-1p-by-RP should we support that.

spec.bs
```diff
@@ -384,30 +395,43 @@ try {
...
-    interaction. This concept is not currently part of [[webauthn-3]], but is on
-    the roadmap for a future enhancement.
+    NOTE: See <a href="https://github.com/w3c/webauthn/issues/1667">WebAuthn
+    issue 1667</a>.
 
-    Until this is available, user agents can consider either storing a local
-    cache of WebAuthn credentials created for this device, or always showing the
-    transaction UX even if the user may be unable to complete the
-    authentication.
+: <dfn>Steps to silently determine if a credential is valid for the current device</dfn>
```
ianbjacobs, line 417:
```suggestion
: <dfn>Steps to silently determine if a credential ID is valid for the current device</dfn>
```

spec.bs
```diff
@@ -384,30 +395,43 @@ try {
...
+:: An as-yet undefined process by which a user agent can, given a [=credential
+    ID=], silently (i.e., without user interaction) determine if the credential
+    represented by that credential ID is valid for the current device (i.e.,
```
ianbjacobs, line 420: Can we use a term other than "valid"?  What about "is available" ?

spec.bs
ianbjacobs: I see we also say further down: "This specification does not preclude their use in other use-cases (e.g., login)." So maybe we only need to say it once (whether here or below).

stephenmcgruer: Done.

spec.bs

spec.bs

ianbjacobs (approved):

Issue #164, opened 2021-12-02
Proposal: add Relying Party id as a required input to SPC authentication


stephenmcgruer:
# Background

Currently, SPC takes as input [a list of credential IDs](https://w3c.github.io/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-credentialids), but does not take the [Relying Party](https://w3c.github.io/webauthn/#webauthn-relying-party) ID as input. Instead, we currently 'magically' obtain it during the "[Steps to respond to a payment request](https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-respond-to-a-payment-request)".

Currently in Chrome, that magic is part of the browser-local storage that we intend to get rid of (long-term). We had previously envisioned that in the future, the magic would be delivered by authenticator-level support, e.g. as part of our assumed use of [a Conditional UI-like API](https://w3c.github.io/secure-payment-confirmation/#webauthn-conditional-ui).

However, in talking with the WebAuthn folks, it has become clear that having the RP ID as an input would be beneficial. Authenticators often internally store discoverable credentials as an `RP id --> [credential]` structure, and so looking up from a credential ID to an RP ID could be quite expensive. Additionally, Conditional UI presumes that the RP ID is known, and so it will be easier to 'piggyback' on that API support if we also have the RP ID as an input.

Having the RP ID as an input would also solve another edge-case problem - what happens if the SPC caller passes in credentials from multiple different RPs? Today this is undefined; the behaviour will be reasonable but inconsistent. By requiring the RP ID as input, it becomes clear - any credential ID from a different RP will simply not match.

# Proposal

Add RP ID as a required member of `SecurePaymentConfirmationRequest`:

```
dictionary SecurePaymentConfirmationRequest {
    required BufferSource challenge;
    required UVString rpId;
    required FrozenArray<BufferSource> credentialIds;
    required PaymentCredentialInstrument instrument;
    unsigned long timeout;
    required USVString payeeOrigin;
    AuthenticationExtensionsClientInputs extensions;
};
```

This RP ID would be a *claim* from the SPC caller as to whom the RP is for the passed credentials. This claim would be verified by the authenticator as part of the flow (as the credential IDs wouldn't match if the wrong RP ID was provided).

## Alternative - optional RP ID

[...]

PR #173, opened 2022-02-03
Require Relying Party ID as input

nickburris:
Add a required input to `SecurePaymentConfirmationRequest` for the
relying party ID. If the requested credentials do not have a matching
RP ID, then the authentication will fail in the "Verifying an
Authentication Assertion" steps.

Before UI is shown, i.e. in "Steps to check if a payment can be made",
the input rpId is only checked to be a valid URL. This ensures that this
is a breaking change before any UI is shown. The actual validation of
the RP ID happens during authentication.

Issue #164

***
<a href="https://pr-preview.s3.amazonaws.com/nickburris/secure-payment-confirmation/pull/173.html" title="Last updated on Feb 4, 2022, 4:17 PM UTC (b9eb7c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/173/edcdaa5...nickburris:b9eb7c7.html" title="Last updated on Feb 4, 2022, 4:17 PM UTC (b9eb7c7)">Diff</a>


spec.bs
```diff
@@ -548,6 +553,11 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is empty,
     return `false`.
 
+1. Let |parsedRpId| be the result of running the [=URL parser=] on
+    |data|["{{SecurePaymentConfirmationRequest/rpId}}"].
+
+1. If |parsedRpId| is failure, then return `false`.
```
stephenmcgruer, line 559: Can you explain these steps further? In the description you say "This ensures that this is a breaking change before any UI is shown.", but I don't quite follow what that means.

spec.bs
```diff
@@ -632,7 +642,7 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     : {{AuthenticationExtensionsPaymentInputs/isPayment}}
     :: The boolean value `true`.
     : {{AuthenticationExtensionsPaymentInputs/rp}}
-    :: **TODO**
+    :: |data|["{{SecurePaymentConfirmationRequest/rpId}}"]
```
stephenmcgruer, line 645: Please remove the TODO on line 655 below.

spec.bs
nickburris: The intention was that if a user (developer) of SPC does not include the rpId input, then SPC fails before any UI is shown, since the rpId is only actually validated during webauthn authentiation after the UI is accepted. Although I now realize this isn't really an issue if it's a required input field, since the SecurePaymentConfirmationRequest creation would fail if rpId isn't included.

This might still be a reasonable preliminary check, e.g. if the developer is dynamically setting rpId and something went wrong and they've set it to a null/empty string, we can fail SPC before UI is shown here. WDYT?

stephenmcgruer: > since the SecurePaymentConfirmationRequest creation would fail if rpId isn't included.

Yep, exactly!

> This might still be a reasonable preliminary check, e.g. if the developer is dynamically setting rpId and something went wrong and they've set it to a null/empty string, we can fail SPC before UI is shown here. WDYT?

So I think we should probably check it here, because WebAuthn actually doesn't as currently spec'd (surprise!). WebAuthn isn't checking it for two reasons:

1. We don't set `options.rpId` in step 4 [of this alg](https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-respond-to-a-payment-request) - this is a bug and we should fix that.
2. We [override](https://w3c.github.io/secure-payment-confirmation/#client-extension-processing-authentication) step 6.1 of [this WebAuthn algorithm](https://w3c.github.io/webauthn/#dom-publickeycredential-discoverfromexternalsource-slot), which removes any check that would happen for `rpId` being a domain.

So I think we should either do that check here, or put the check into our extension override. In either case, we need to do more than just parse the URL - we need to make sure its a specifically [valid domain](https://url.spec.whatwg.org/#valid-domain). I think we can just reference that URL spec algorithm to do so, it seems to take a string as input and check that its a domain?

nickburris: IIUC the rpId is already validated as part of this spec in [Verifying an Authentication Assertion](https://w3c.github.io/secure-payment-confirmation/#sctn-verifying-assertion), step 2.5 _"After step 13, insert the following steps: ... Verify that the value of C["[payment](https://w3c.github.io/secure-payment-confirmation/#dom-collectedclientpaymentdata-payment)"]["[rp](https://w3c.github.io/secure-payment-confirmation/#dom-collectedclientadditionalpaymentdata-rp)"] matches the [Relying Party](https://w3c.github.io/webauthn/#relying-party)'s origin."_. Doesn't verifying that it matches effectively do the domain check? Or should we also make sure it's a valid domain here?

stephenmcgruer: Those checks are for the Relying Party to make on their own server, after the assertion has been generated. They're not part of any browser (or authenticator) algorithm.

stephenmcgruer: Also, I said above 'make the check here or in the authenticator algorithm override', but of course we want it to happen before any UI is shown so it should go here.

spec.bs
nickburris: Done

spec.bs
```diff
@@ -548,6 +553,9 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is empty,
     return `false`.
 
+1. If |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is not a
+    [valid domain](https://url.spec.whatwg.org/#valid-domain), return `false`.
```
stephenmcgruer, line 557: Normally I would say to do this via an [anchor block](https://tabatkins.github.io/bikeshed/#custom-dfns) (e.g. see our existing block on line 24), but interestingly enough it looks like WebAuthn does it via a [link-default](https://tabatkins.github.io/bikeshed/#link-defaults) block instead...

So let's [copy WebAuthn](https://github.com/w3c/webauthn/blob/main/index.bs#L313) for now, and we can ask Dom to take a look at the PR when ready.

spec.bs
stephenmcgruer: > Done. So IIUC this means webauthn will now validate that rpId matches the credential's RP.

Yep!

> Done, although #valid-domain isn't something that's exported from the url spec, so I have to link to it directly, since [[url#valid-domain]] doesn't work. Do you think that's okay to do?

See my comment on that line.

spec.bs
nickburris: Cool! Done.

stephenmcgruer (approved):
LGTM modulo one thing I forgot - can you update the example code in section 1.2.2?

After that, we should again leave this open for (say) a week for comments, since this will be a breaking change (once implemented in a browser). In particular, cc @jcemer-stripe, and... I don't have a GitHub username handy for our Adyen friends :/. @ianbjacobs , do you have one?

spec.bs
```diff
@@ -358,6 +362,8 @@ const request = new PaymentRequest([{
     // List of credential IDs obtained from the bank.
     credentialIds,
 
+    rpId: "https://fancybank.com/",
```
stephenmcgruer, line 365: If `rpId` is a domain, should this be just `fancybank.com`? I'm not too familiar with WebAuthn, so might be worth testing this out and seeing!

nickburris: Ah yep, it should just be fancybank.com (confirmed with webauthn enrollment, at least). Fixed.

PR #178, opened 2022-03-02
[Spec][Draft] Require user-activation for credential creation

### LINKED ISSUE #128: Clarification on requiring user activation for credential creation in cross-origin iframes to mitigation cross-site tracking
https://github.com/w3c/secure-payment-confirmation/blob/main/explainer.md#enrollment-in-cross-origin-iframes mentions that "if site A embeds an iframe from site B, with the "payment" policy set, then site B may initiate a credential creation for site B within that iframe", with a proposed mitigation of requiring transient user activation. 

For secure payment confirmation specifically, would the credential creation only happen after the browser payment UX flow is completed by the user? The browser payment UX seems stronger than mere user activation, since it's controlled by the browser and can't be faked (for instance, a page couldn't "for example - overlay a transparent iframe on-top of a legitimate button" on top of the payment UX for the purpose of cross-site tracking). Is there something I'm missing?

stephenmcgruer:
Fixes #128

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -793,6 +793,21 @@ directly; for authentication the extension can only be accessed via
 
     When [[webauthn-3#sctn-createCredential|creating a new credential]]:
 
+    1. Before step 2, insert two new steps:
+
+        * If the [=relevant global object=], as determined by the calling
+            {{CredentialsContainer/create()}} implementation, does not have
+            [=transient activation=]:
+
+             * Return a {{DOMException}} whose name is "{{SecurityError}}", and
+                terminate this algorithm.
+
+        * [=Consume user activation=] of the [=relevant global object=].
+
+         Note: These steps aim to protect the user against malicious iframes
+         attempting to trick the user into registering. See
+         [[#sctn-privacy-user-activation]].
+
     1. Remove the check for *sameOriginWithAncestors* in step 2.
 
          Note: This allows for creating SPC credentials in a cross-origin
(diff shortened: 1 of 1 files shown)


nickburris (commented):
Github won't let me comment on the exact line, but there's a paragraph in the Privacy Considerations section 11.1 that says:

> A consideration for a future version of this specification might be to
> additionally require consuming a [=transient activation=] for credential
> creation in a cross-origin iframe, to ensure that the user has at least
> interacted with an iframe attempting credential creation.

Let's update it to say something like:

> This attack is further mitigated by requiring the consumption
> of a [=transient activation=] for credential creation in a cross-origin
> iframe, to ensure that the user has at least interacted with an iframe
> attempting credential creation."

stephenmcgruer:
Good catch; I totally forgot we already had a section on this! Fixed.

stephenmcgruer:
After today's discussion with WebAuthn folks, I've changed this PR to only require a user activation when creating a credential in a cross-origin iframe. This better aligns with what this would look like if WebAuthn started to allow credential creation in a cross-origin iframe.

spec.bs
```diff
@@ -1293,46 +1304,18 @@ NOTE: Allowing credential creation in cross-origin iframes is currently [under
...
-A consideration for a future version of this specification might be to
-additionally require consuming a [=transient activation=] for credential
-creation in a cross-origin iframe, to ensure that the user has at least
-interacted with an iframe attempting credential creation.
+Allowing credential creation in a cross-origin iframe presents a risk that an
+iframe may attempt to trick a user into registering a credential. That
+credential could then be used for tracking (see [WebAuthn issue
+1336](https://github.com/w3c/webauthn/issues/1336#issue-513568855)). To
+mitigate such an attack, this specification requires that a call to
+{{CredentialsContainer/create()|navigator.credentials.create()}} with the
```
nickburris, line 1312: (did my comment disappear?)

Nit: consider making this more explicit; "... `navigator.credentials.create()` inside of a cross-origin iframe with the..."

stephenmcgruer: Good catch! Fixed.

PR #179, opened 2022-03-02
[Spec] Make example clearer about iframe interactions

### LINKED ISSUE #168: [Spec] Registration example is misleading around whether an iframe is required
Minor issue that came up when discussing SPC with someone.

In [1.2.1. Registration](https://w3c.github.io/secure-payment-confirmation/#sctn-sample-registration), we describe an example flow where the user registers in a cross-origin iframe during a transaction. A number of issues here has led to confusion aroudn whether the cross-origin iframe is *required*:

1. We refer to it as '_the_ first-time flow' (emphasis mine), rather than 'a first-time flow' or other more specific wording.
2. We say 'The sample code for registering the user follows', again rather than being more specific about this being one option for registration. (E.g. 'for registering the user in this way')
3. The example code documents the `"payment"` extension as 'required to allow credential creation in an iframe, and so that the browser knows this credential relates to SPC' - these should probably be swapped to indicate which half is more important.

I think we should fix these up at some point, to help aid understanding of SPC. We should perhaps also change [1.1.2. Registration in a third-party iframe](https://w3c.github.io/secure-payment-confirmation/#sctn-use-case-iframe-registration) to just be 'Registration', and then have 'in a third-party iframe' as a sub-point? I'm less confident in that though!

stephenmcgruer:
Fixes #168

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/179.html" title="Last updated on Mar 2, 2022, 5:39 PM UTC (470e48c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/179/2d0c9a2...470e48c.html" title="Last updated on Mar 2, 2022, 5:39 PM UTC (470e48c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -220,7 +220,7 @@ example flows and do not limit the scope of how the API can be used.
 
 ### Registration ### {#sctn-sample-registration}
 
-This is the first-time flow, in which a new credential is created and stored by
+This is a first-time flow, in which a new credential is created and stored by
 an issuing bank.
 
 1. The user visits `merchant.com`, selects an item to purchase, and proceeds to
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
cc @agektmr as an fyi

ianbjacobs (approved):

Issue #142, opened 2021-10-07
11.2 define the mitigation


samuelweiler:
The attack in 10.2 is clear, as is the goal ("make sure not to enable malicious callers... to distinguish between these cases").

But the methods of getting there are not defined.  Instead of "If the URL is accessed then the caller can conclude that at least one of the passed credentials is available to the user", how about an instruction: "do not load the URL until"?  Presumably that instruction should be elsewhere in the doc, to be read contemporaneously with other process test.  Are any other mitigations needed to achieve the goal?

stephenmcgruer:
(Note: this is now [11.2](https://w3c.github.io/secure-payment-confirmation/#sctn-privacy-probing-credential-ids))

samuelweiler:
I'm glad to see the explicit instructions in what is now 4.1.4.  

1) I'd like 11.2 to point at 4.1.4 (e.g. "here's now to achieve this"

2) Are the instructions in 4.1.4 sufficient to mitigate all leaks?  From your text, I'm inferring that is it not.  What else is needed?

ianbjacobs:
Thank you, @samuelweiler, for the recommendation to tighten up the normative language, now integrated. We hope this resolves the current issue.

PR #182, opened 2022-04-27
[Spec] Be clearer about authentication ceremony privacy in 4.1.4

stephenmcgruer:
Before this PR, the spec was vague about the timing attack and also didn't
specify at all that an implementation must return a NotAllowedError instead of
the normal PaymentRequest NotSupportedError in the case of
no-matching-credentials. Whilst we don't want to enforce that UAs show a dialog
here (they may decide, for example, that a delayed response instead is
sufficiently privacy preserving), this PR does try to make the actual concern
clearer and more normative, and add a normative requirement for the return
value.

Should also improve the situation called out in
https://github.com/w3c/secure-payment-confirmation/issues/142

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/182.html" title="Last updated on Apr 27, 2022, 2:01 PM UTC (ffd6cf7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/182/bf15f84...ffd6cf7.html" title="Last updated on Apr 27, 2022, 2:01 PM UTC (ffd6cf7)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -623,11 +623,17 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
         |id| from |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"].
 
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is now empty,
-    return `false`. The user agent must take care to maintain
-    [[#sctn-privacy-probing-credential-ids|authentication ceremony privacy]],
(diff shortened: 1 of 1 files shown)


ianbjacobs (approved):

rsolomakhin (approved):
Looks good, thank you.

PR #195, opened 2022-06-30
[Spec] Move some input validation to construction time

### LINKED ISSUE #194: SPC spec should validate some parameters in the constructor rather than steps-to-check-if-a-payment-can-be-made
Currently, the spec does all its data validation in the [steps to check if a payment can be made](https://w3c.github.io/secure-payment-confirmation/#sctn-steps-to-check-if-a-payment-can-be-made), which are called [during show() processing](https://w3c.github.io/payment-request/#show-method) (step 16.6.1). This is at odds with the actual implementation in Chrome, which validates a lot of the data during **construction**, not `show()` (e.g., see https://wpt.fyi/results/secure-payment-confirmation/constructor.https.html?label=master&label=experimental&aligned).

I believe that validating the data during construction is actually correct (or at least, useful) as it is trivial checking (e.g., `If data["credentialIds"] is empty, return false.`). There is actually a PaymentRequest issue tracking adding a concept for this - https://github.com/w3c/payment-request/pull/976, for which I have a PR (https://github.com/w3c/payment-request/pull/977) that could address it, but it needs a little TLC.

I would like us to try to get that PR landed, and then update the SPC spec to do some of the validation during construction.

stephenmcgruer:
Fixes #194

Depends on https://github.com/w3c/payment-request/pull/977

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/195.html" title="Last updated on Jul 7, 2022, 5:07 PM UTC (a314305)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/195/12f96d5...a314305.html" title="Last updated on Jul 7, 2022, 5:07 PM UTC (a314305)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -607,8 +607,18 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is empty,
     throw a {{RangeError}}.
 
+1. For each |id| in |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"]:
+
+    1. If |id| is empty, throw a {{RangeError}}.
+
 1. If |data|["{{SecurePaymentConfirmationRequest/challenge}}"] is null or
     empty, throw a {{TypeError}}.
+
+1. If |data|["{{SecurePaymentConfirmationRequest/instrument}}"]["{{PaymentCredentialInstrument/displayName}}"]
+    is empty, throw a {{TypeError}}.
+
+1. If |data|["{{SecurePaymentConfirmationRequest/instrument}}"]["{{PaymentCredentialInstrument/icon}}"]
+    is empty, throw a {{TypeError}}.
 
 1. If |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is not a
     [=valid domain=], throw a {{TypeError}}.
@@ -629,9 +639,6 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
     1. If |parsedURL| is failure, then throw a {{TypeError}}.
 
     1. If |parsedURL|'s [=url/scheme=] is not "`https`", then throw a {{TypeError}}.
-
-1. If |data|["{{SecurePaymentConfirmationRequest/instrument}}"]["{{PaymentCredentialInstrument/displayName}}"]
-    is empty, throw a {{TypeError}}.
 
 ### Steps to check if a payment can be made ### {#sctn-steps-to-check-if-a-payment-can-be-made}
 
--- a/spec.bs
+++ b/spec.bs
@@ -588,21 +588,28 @@ members:
         automatically.
 </dl>
 
-### Steps to check if a payment can be made ### {#sctn-steps-to-check-if-a-payment-can-be-made}
+### Payment Method additional data type ### {#sctn-payment-method-additional-data-type}
 
-The [=steps to check if a payment can be made=] for this payment method, for an
+The `payment method additional data type` for this payment method is {{SecurePaymentConfirmationRequest}}.
+
+NOTE: **TODO**: Link to PaymentRequest spec concept before landing this PR.
+
+### Steps to validate payment method data ### {#sctn-steps-to-validate-payment-method-data}
+
+The `steps to validate payment method data` for this payment method, for an
 input {{SecurePaymentConfirmationRequest}} |data|, are:
+
+NOTE: **TODO**: Link to PaymentRequest spec concept before landing this PR.
 
 <wpt>
   constructor.https.html
-  authentication-invalid-icon.https.html
-  authentication-icon-data-url.https.html
-  authentication-rejected.https.html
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
@rsolomakhin - PTAL, and in particular check that this matches what we check synchronously at construction time versus asynchronously for later canMakePayment()/show() calls.

spec.bs
```diff
@@ -588,48 +590,77 @@ members:
...
 
 <wpt>
   constructor.https.html
-  authentication-invalid-icon.https.html
-  authentication-icon-data-url.https.html
-  authentication-rejected.https.html
 </wpt>
 
-
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is empty,
```
rsolomakhin, line 607: Did you mean to sync up with this [Chrome code](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/payments/secure_payment_confirmation_helper.cc;l=40-101;drc=cfd951a304bb7a1e6424c614c094f5bc581ca8b7)? That code should be executing in the `PaymentRequest()` constructor, as far as I can see. Its sequence of checks is:

1. Credential ID list: non-empty.
2. Timeout: under 1 hour.
3. Instrument display name: non-empty.
4. Instrument icon: valid URL.
5. Either payee origin or name: non-empty.
6. RP ID: non-empty.

This spec here appears to have:

1. Credential ID list: non-empty. ✅
2. Timeout: no checks. (Spec may want to add this.) ⬅️
3. Instrument icon: no checks. (Spec should check this.) 🔥
4. Challenge: non-empty. (Chrome should start doing this.) 🔥
5. RP ID: valid domain. (Chrome should start doing this.) 🔥
6. Either payee name or origin: non-empty. ✅
7. Payee origin: valid https URL. (Chrome should start doing this.) 🔥
8. Instrument display name: non-empty. (Spec may want to move this up to position 3.) ⬅️

There are some differences here, including ordering of checks. Am I looking at the wrong chunk of Chrome code by accident?

spec.bs
```diff
@@ -588,48 +590,77 @@ members:
...
+input {{SecurePaymentConfirmationRequest}} |data|, are:
+
+<wpt>
+  constructor.https.html
+  authentication-invalid-icon.https.html
+  authentication-icon-data-url.https.html
+  authentication-rejected.https.html
+</wpt>
+
+1. If |data|["{{SecurePaymentConfirmationRequest/payeeOrigin}}"] is present:
```
rsolomakhin, line 648: Comparing to the [Chrome code](https://source.chromium.org/chromium/chromium/src/+/main:components/payments/content/secure_payment_confirmation_app_factory.cc;l=44-89;drc=cfd951a304bb7a1e6424c614c094f5bc581ca8b7) that determines whether `canMakePayment()` will return true:

1. Credential IDs list: non-empty. (Fails only on compromised renderer.) ✅
2. Each credential ID: non-empty. ([The `PaymentRequest()` constructor](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/payments/secure_payment_confirmation_helper.cc;l=40-101;drc=cfd951a304bb7a1e6424c614c094f5bc581ca8b7) should be checking this, too.) 🔥
3. Timeout: under 1 hour. (Fails only on compromised renderer.) ✅
4. Challenge: non-empty. ([The `PaymentRequest()` constructor](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/payments/secure_payment_confirmation_helper.cc;l=40-101;drc=cfd951a304bb7a1e6424c614c094f5bc581ca8b7) should be checking this, too.) 🔥
5. Instrument display name: non-empty. (Fails only on compromised renderer.) ✅
6. Instrument icon: valid URL. (Fails only on compromised renderer.) ✅
7. Download the icon.
8. Check for instrument IDs on device.

This spec here has:

1. Payee origin: valid https URL. (Chrome should check that either payeer origin is a valid https URL or payee name is non empty in [here](https://source.chromium.org/chromium/chromium/src/+/main:components/payments/content/secure_payment_confirmation_app_factory.cc;l=44-89;drc=cfd951a304bb7a1e6424c614c094f5bc581ca8b7).) 🔥
2. Download the icon. ✅
3. Check for instrument IDs on device. ✅

rsolomakhin (changes requested):
HI @stephenmcgruer, the code and the spec seem to disagree a bit. Should we get the spec into a more sensible place (re-order some checks, add some more checks here and there) and then open a bug in Chrome side to match the spec better (mostly adding more checks)?

rsolomakhin (approved):
Thank you for making this spec more awesome 😆

This spec change LGTM and feel free to assign the Chrome bug report to me, if you need a helping hand. 👋

stephenmcgruer:
Opened https://crbug.com/1342686 for Chrome changes, and will also send a WPT change to align tests with spec.

Issue #191, opened 2022-06-09
rp vs. rpId


rlin1:
We use rp in CollectedClientAdditionalPaymentData
But we use rpId in SecurePaymentConfirmationRequest

type is USVString in both cases and the meaning seems to be identical, i.e. RP ID.

Why is that?

stephenmcgruer:
I believe this is just an oversight / historical artifact from the early days of the spec.

[CollectedClientAdditionalPaymentData](https://w3c.github.io/secure-payment-confirmation/#dictdef-collectedclientadditionalpaymentdata) came first, and I think used `rp` 'just because', i.e. without any reason.

Adding the RP info to [SecurePaymentConfirmationRequest](https://w3c.github.io/secure-payment-confirmation/#dictdef-securepaymentconfirmationrequest) came later, and by that time we were aware that WebAuthn's `rp` concept is actually a full type, [PublicKeyCredentialRpEntity](https://w3c.github.io/webauthn/#dictdef-publickeycredentialrpentity), and when the WebAuthn spec only needs to encode the ID (e.g. in [PublicKeyCredentialRequestOptions](https://w3c.github.io/webauthn/#dictionary-assertion-options)), it names it `rpId`. So we followed that.

I believe we should align on the latter (`rpId`) in both these cases. We can either do that today with a backwards compatible change (e.g., start making `CollectedClientAdditionalPaymentData` return an `rpId` member, but note that implementations may **also** return `rp` due to historical reasons), or we can do that if/when we do a full API change away from PaymentRequest (tracked by #65 and #56 , approximately). I'm happy with either :).

stephenmcgruer:
This has been fixed in the specs, tests, and [in Chrome](https://chromium-review.googlesource.com/c/chromium/src/+/3854551). On the Chrome side, once the change is in Stable (target end of October) we will send out an announcement asking folks to switch their validation logic over, with some deprecation period TBD (probably 3-6 months).

PR #198, opened 2022-08-10
[Spec] Rename rp --> rpId in CollectedClientAdditionalPaymentData

stephenmcgruer:
To align with WebAuthn, we should use the term rpId here. This is a breaking
change, but implementations can mitigate the breakage by continuing to include
the old 'rp' name going forwards.

See https://github.com/w3c/secure-payment-confirmation/issues/191

Test changes: https://github.com/web-platform-tests/wpt/pull/35602
Implementation bugs:
- Chrome: https://crbug.com/1356224
- Safari: N/A
- Firefox: N/A

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/198.html" title="Last updated on Aug 24, 2022, 3:49 PM UTC (dd14bfa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/198/7204dc0...dd14bfa.html" title="Last updated on Aug 24, 2022, 3:49 PM UTC (dd14bfa)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -770,7 +770,7 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
 
     : {{AuthenticationExtensionsPaymentInputs/isPayment}}
     :: The boolean value `true`.
-    : {{AuthenticationExtensionsPaymentInputs/rp}}
+    : {{AuthenticationExtensionsPaymentInputs/rpId}}
     :: |data|["{{SecurePaymentConfirmationRequest/rpId}}"]
     : {{AuthenticationExtensionsPaymentInputs/topOrigin}}
     :: |topOrigin|
@@ -872,7 +872,7 @@ directly; for authentication the extension can only be accessed via
       boolean isPayment;
 
       // Only used for authentication.
-      USVString rp;
+      USVString rpId;
       USVString topOrigin;
       DOMString payeeName;
       USVString payeeOrigin;
@@ -887,7 +887,7 @@ directly; for authentication the extension can only be accessed via
 
         <div class="note">**TODO**: Find a better way to do this. Needed currently because other members are auth-time only.</div>
 
-      :  <dfn>rp</dfn> member
+      :  <dfn>rpId</dfn> member
       :: The [=Relying Party=] id of the credential(s) being used. Only used at authentication time; not registration.
 
       :  <dfn>topOrigin</dfn> member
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
This probably warrants discussion in the WPWG, or at least some communication to anyone who may be validating an SPC credential today. On the Chrome side, we would likely have to keep the `rp` version of the entry around for now, and eventually do a deprecation process to switch over to rpId.

stephenmcgruer:
Worth noting that it will be actually impossible for us to know if anyone is still relying on `rp` being included in the assertion (at whatever future point in time we did a deprecation), as it will usually be verified server-side.

ianbjacobs:
I will add to the 18 August WPWG agenda.

ianbjacobs:
cc @stare893

ianbjacobs (approved):

ianbjacobs:
See resolution at the 18 August meeting to make this change:
https://www.w3.org/2022/08/18-wpwg-minutes.html#t01

stare893:
@stephenmcgruer There is some confusion on the proposed change. From EMV 3DSWG consideration, we already see rpId in use per the SPC API webpage and we have included the same in all 3DS spec references to the SPC API use. Can you please elaborate/point me in the right direction to understand the change.

@ianbjacobs  you might get a similar query through the 3dswg official email

ianbjacobs:
@stare893 You are correct that SPC input refers to the Web Authentication rpid; see the request dictionary:
 https://w3c.github.io/secure-payment-confirmation/#sctn-securepaymentconfirmationrequest-dictionary

This proposal is to change the field name in the output; see the additional payment dictionary:
 https://w3c.github.io/secure-payment-confirmation/#sctn-collectedclientadditionalpaymentdata-dictionary

The proposal is that assertions will include the field name "rpid" to align with the Web Authentication name. 

The Chrome implementation would support both "rp" and "rpid" field names for some period of time, eventually deprecating "rp".

It was not clear to us on the call today whether the 3DS specification itself would need to change (e.g., because it refers to the "rp" field in the resulting assertion), or if instead it would just be implementations (e.g., ACS) that would need to adapt code to look for "rpid" rather than "rp" in the assertion.

ianbjacobs:
@stephenmcgruer, EMVCo folks have confirmed that this change does not impact their 3DS integration (the spec itself). Given the WG's support, let's merge this.

Issue #219, opened 2022-12-15
Use lowercase values in enum


dontcallmedom:
#215 introduced enum values that use camel-case; the convention for Web specs is to use [lowercase, dash-separated values for enums](https://w3ctag.github.io/design-principles/#casing-rules)

rsolomakhin:
Should it be "auto-accept", "auto-reject", "auto-opt-out"?

stephenmcgruer:
Interesting; we were advised to switch *to* camelcase by our internal WebDriver (well, ChromeDriver) team, to align with other such commands. I don't know if the specific person who recommended that interacts on GitHub much, but perhaps @foolip can suggest someone with WebDriver spec experience who can weight in on how enums for WebDriver should be cased?

dontcallmedom:
Oh,  I had missed this was used in webdriver context. I'm not sure using webidl to describe a list of possible string values in webdriver makes sense in the first place...

dontcallmedom:
>  I'm not sure using webidl to describe a list of possible string values in webdriver makes sense in the first place

While noting that I'm not at all a WebDriver expert, looking at the WebDriver specification itself, these lists of possible values tend to be described as keyword tables, binding keywords with a state and a description. I think a likely reasonable match to what SPC is trying to do here is the [known prompt handling approaches table ](https://w3c.github.io/webdriver/#dfn-known-prompt-handling-approaches-table).

I'll further note that while names for keys in the JSON objects used in WebDriver tend to use camelCase, the string values don't seem to be particularly using that convention.

foolip:
There aren't many WebDriver experts in the world, and wouldn't count myself amongst them, but I'll just go ahead and read https://w3c.github.io/webdriver/ and look for precedent.

Capabilities are camelCase: https://w3c.github.io/webdriver/#capabilities

Locator strategies are lowercase with spaces: https://w3c.github.io/webdriver/#locator-strategies

The Actions API uses camelCase: https://w3c.github.io/webdriver/#actions

https://w3c.github.io/permissions/#automation will use some lowercase-dashed identifiers from https://w3c.github.io/permissions-registry/, which are also used in JS APIs I think.

So my two suggestions would be:
- Don't use Web IDL to define this, unless that enum is also used in some JS API and you do something similar to permissions.
- Use camelCase if these strings will only appear in the WebDriver protocol.

But if @sadym-chromium or @nechaev-chromium say something different, then ignore me and listen to them :)

[...]

PR #231, opened 2023-02-16
Use an hyphenated version of the SPC Transaction Mode enumeration values

javifernandez:
According to the [Web Platform Design Principles ](https://w3ctag.github.io/design-principles/#casing-rules) the enumeration values should be defined in the "Lowercase, dash-delimited" form. 

I've recently defined a new Web Driver command, inspired by the [Set SPC Transaction Mode](https://w3c.github.io/secure-payment-confirmation/#set-spc-transaction-mode) command defined in this spec, for the Custom Handler feature. The new  [Set RPH Registration Mode](https://html.spec.whatwg.org/multipage/system-state.html#set-rph-registration-mode) command values are defined according to the mentioned rules: auto-accept and auto-reject. 

Is there any reason or specific need in the spec to avoid following the design principle casting rules ?  Would make sense to have a common solution for both web driver commands.

As a mater of fact, I've recently landed a [CL](https://chromium-review.googlesource.com/c/chromium/src/+/4218519) for the Chrome Dev Tools so that both command's implementation for Chromedriver use the same enum. This shared enum still uses the camelCase version, so it'd should be a matter of changing it to use the hyphenated version instead.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -792,15 +792,15 @@ should then bypass the above communication of information and gathering of user
 consent, and instead do the following based on the value of the [=current
 transaction automation mode=]:
 
-    : "`autoAccept`"
+    : "`auto-accept`"
     :: Act as if the user has seen the transaction details and accepted
         the authentication.
 
-    : "`autoReject`"
+    : "`auto-reject`"
     :: Act as if the user has seen the transaction details and rejected
         the authentication.
 
-    : "`autoOptOut`"
+    : "`auto-opt-out`"
     :: Act as if the user has seen the transaction details and indicated
         they want to opt out.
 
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
Hi there :)

This was recently discussed in https://github.com/w3c/secure-payment-confirmation/issues/219, and the conclusion was that for webdriver parameters, camelCase was correct. Please give that issue a read through, and if you still disagree then maybe we can re-open it and discuss there (or maybe it becomes a topic for the webdriver spec itself)

ianbjacobs:
@javifernandez are you ok if we close this issue with no changes?

javifernandez:
> @javifernandez are you ok if we close this issue with no changes?

I'm not sure we have got the consensus we aimed for in the [WebDriver related discussion](https://github.com/w3c/webdriver/issues/1725#issuecomment-1492400430). It seems that we all agree on implementing whatever is decided; @annevk is also open to accept any proposal from the WebDriver people. I guess we would just need  some confirmation there.

PR #230, opened 2023-02-09
Note in explainer re: user activation (removal)

ianbjacobs:
Per TAG review [1], adding a note to Security Considerations about the click-jacking risk and how the browser mitigates it.

[1] https://github.com/w3ctag/design-reviews/issues/802#issuecomment-1422225875

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/230.html" title="Last updated on Apr 24, 2023, 2:58 PM UTC (f323fbe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/230/1565e06...f323fbe.html" title="Last updated on Apr 24, 2023, 2:58 PM UTC (f323fbe)">Diff</a>

PR #233, opened 2023-03-09
Add isSecurePaymentConfirmationAvailable API to spec

### LINKED ISSUE #81: Make it easy to determine whether a user agent supports SPC

nickburris:
This adds spec text for an API for developers to more easily check whether SPC is available. Fixes #81. One question is whether the spec should further clarify what it means for SPC to be available, but I'm not sure how we can say that without being too implementation specific.

***
<a href="https://pr-preview.s3.amazonaws.com/nickburris/secure-payment-confirmation/pull/233.html" title="Last updated on Apr 11, 2023, 7:31 PM UTC (484e07a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/233/568fbab...nickburris:484e07a.html" title="Last updated on Apr 11, 2023, 7:31 PM UTC (484e07a)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -637,8 +637,8 @@ partial interface PaymentRequest {
 <dl dfn-type="attribute" dfn-for="PaymentRequest">
     :   {{PaymentRequest/isSecurePaymentConfirmationAvailable()}}
     ::  Upon invocation, a promise is returned that resolves with a value of
-        `true` if the user agent believes that the Secure Payment Confirmation
-        feature is available, or `false` otherwise.
+        `true` if the Secure Payment Confirmation feature is available, or
+        `false` otherwise.
 </dl>
 
 This allows a developer to perform the following check when deciding whether to
--- a/spec.bs
+++ b/spec.bs
@@ -624,7 +624,7 @@ The [=payment method additional data type=] for this payment method is
 
 ### Checking if Secure Payment Confirmation is available ### {#sctn-secure-payment-confirmation-available-api}
 
-A static API is added to {{PaymentRequest}} in order to provide develoeprs a
+A static API is added to {{PaymentRequest}} in order to provide developers a
 simplified method of checking whether Secure Payment Confirmation is available.
 
 <xmp class="idl">
--- a/spec.bs
+++ b/spec.bs
@@ -642,7 +642,7 @@ partial interface PaymentRequest {
 This allows a developer to perform the following check when deciding whether to
 initiate a SPC flow:
 
-<pre class="example" id="registration-example" highlight="js">
+<pre class="example" id="available-example" highlight="js">
 const spcAvailable =
     PaymentRequest &&
     PaymentRequest.isSecurePaymentConfirmationAvailable &&
--- a/spec.bs
+++ b/spec.bs
@@ -420,10 +420,6 @@ try {
 }
 </pre>
 
-NOTE: The use of the static {{PaymentRequest/isSecurePaymentConfirmationAvailable}} method is recommended for
-      SPC feature detection, instead of calling {{PaymentRequest/canMakePayment}} on an already-constructed
-      PaymentRequest object.
-
 # Terminology # {#sctn-terminology}
 
 : <dfn export>SPC Credential</dfn>
@@ -654,6 +650,10 @@ const spcAvailable =
     PaymentRequest.isSecurePaymentConfirmationAvailable &&
     await PaymentRequest.isSecurePaymentConfirmationAvailable();
 </pre>
+
+NOTE: The use of the static {{PaymentRequest/isSecurePaymentConfirmationAvailable}} method is recommended for
+      SPC feature detection, instead of calling {{PaymentRequest/canMakePayment}} on an already-constructed
+      PaymentRequest object.
 
 ### Steps to validate payment method data ### {#sctn-steps-to-validate-payment-method-data}
 
--- a/spec.bs
+++ b/spec.bs
@@ -363,7 +363,16 @@ The sample code for authenticating the user follows. Note that the example code
 presumes access to await/async, for easier to read promise handling.
 
 <pre class="example" id="authentication-example" highlight="js">
-if (!window.PaymentRequest) { /* PaymentRequest not available; merchant should fallback to traditional flows */ }
+/* isSecurePaymentConfirmationAvailable indicates whether the browser */
+/* supports SPC. It does not indicate whether the user has a credential */
+/* ready to go on this device. */
+const spcAvailable =
+  PaymentRequest &&
+  PaymentRequest.isSecurePaymentConfirmationAvailable &&
+  await PaymentRequest.isSecurePaymentConfirmationAvailable();
+if (!spcAvailable) {
+  /* Browser does not support SPC; merchant should fallback to traditional flows. */
+}
 
 const request = new PaymentRequest([{
   supportedMethods: "secure-payment-confirmation",
@@ -398,17 +407,7 @@ const request = new PaymentRequest([{
     },
   });
 
-  /* isSecurePaymentConfirmationAvailable indicates whether the browser */
-  /* supports SPC. It does not indicate whether the user has a credential */
-  /* ready to go on this device. */
-
 try {
-  const spcAvailable =
-    PaymentRequest &&
-    PaymentRequest.isSecurePaymentConfirmationAvailable &&
-    await PaymentRequest.isSecurePaymentConfirmationAvailable();
-  if (!spcAvailable) { /* Browser does not support SPC. Handle error. */ }
-
   const response = await request.show();
   await response.complete('success');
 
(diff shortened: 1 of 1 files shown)


ianbjacobs:
Thanks @nickburris and @stephenmcgruer, 

I don't have a strong view; let's try to get input from users of the API. 

Having said that, I see some small advantages to isSecurePaymentConfirmationAvailable:

1) isSecurePaymentConfirmationAvailable appears to be more straightforward in meaning than canMakePayment.
2) This new approach might ease a transition away from SPC-via-Payment-Request if we choose that path.

stephenmcgruer:
> I don't have a strong view; let's try to get input from users of the API.

How/when do you want to gather this input?

> Having said that, I see some small advantages to isSecurePaymentConfirmationAvailable: [...]

Yes, I think we definitely want to make isSecurePaymentConfirmationAvailable the primary way to do availability-checking for the API. The question is more "what about canMakePayment" in such a world - should the SPC spec try to address it (either by defining it to do 'something' useful, TBD what, or by explicitly saying not to use it for SPC)?

ianbjacobs:
> > I don't have a strong view; let's try to get input from users of the API.
> 
> How/when do you want to gather this input?

I have reached out to Stripe and Adyen colleagues for input. I'll also add this to our next meeting agenda.

> 
> > Having said that, I see some small advantages to isSecurePaymentConfirmationAvailable: [...]
> 
> Yes, I think we definitely want to make isSecurePaymentConfirmationAvailable the primary way to do availability-checking for the API. The question is more "what about canMakePayment" in such a world - should the SPC spec try to address it (either by defining it to do 'something' useful, TBD what, or by explicitly saying not to use it for SPC)?

While SPC is grounded in PR API I don't see a reason to try to get rid of canMakePayment. So would this make sense language for users of the API:

* Use isSecurePaymentConfirmationAvailable
* canMakePayment is the old way of doing this; this approach is deprecated.
* Implementations will continue to support both (until the mythical day when SPC is no longer grounded in PR API, at which point canMakePayment could naturally fall away).

stephenmcgruer:
> I have reached out to Stripe and Adyen colleagues for input. I'll also add this to our next meeting agenda.

Ack. Note that we would like to ship this in M113 (branches March 23rd, for stable release in early May), so we are likely to pursue that anyway as we don't foresee any objections from the WG. If for some reason there is significant push-back at the remote meeting (week of March 27th), we should be fine to halt the brakes that close post-branch.

> While SPC is grounded in PR API I don't see a reason to try to get rid of canMakePayment. So would this make sense language for users of the API:

Yep, sgtm. Practically, what I believe this amounts to is:

1. Spec isSecurePaymentConfirmationAvailable
2. Add a spec note saying "canMakePayment also exists, but we recommend using isSecurePaymentConfirmationAvailable instead"

Does that SGTY @ianbjacobs ?

omertoast:
> 1. Spec isSecurePaymentConfirmationAvailable
> 2. Add a spec note saying "canMakePayment also exists, but we recommend using isSecurePaymentConfirmationAvailable instead"

SGTM. As someone who didn't know this issue exist, it seems straightforward and simple as it gets.

adrianhopebailie:
> until the mythical day when SPC is no longer grounded in PR API

`< sigh />`

nickburris:
> Can you update the example in section 1.2.2 to: (1) use your `const spcAvailable = ...`, and to (2) remove or update the comment about canMakePayment ?

Done.

spec.bs
```diff
@@ -398,13 +398,16 @@ const request = new PaymentRequest([{
...
-  /* canMakePayment indicates whether the browser supports SPC. */
-  /* canMakePayment does not indicate whether the user has a credential */
+  /* isSecurePaymentConfirmationAvailable indicates whether the browser */
+  /* supports SPC. It does not indicate whether the user has a credential */
   /* ready to go on this device. */
 
 try {
-  const canMakePayment = await request.canMakePayment();
-  if (!canMakePayment) { /* Browser does not support SPC. Handle error. */ }
+  const spcAvailable =
```
stephenmcgruer, line 406: Optional: I wonder if we should move this up to line 366, where there is already a feature detection check?

spec.bs
```diff
@@ -622,6 +625,33 @@ members:
...
+
+This allows a developer to perform the following check when deciding whether to
+initiate a SPC flow:
+
+<pre class="example" id="available-example" highlight="js">
+const spcAvailable =
+    PaymentRequest &&
+    PaymentRequest.isSecurePaymentConfirmationAvailable &&
+    await PaymentRequest.isSecurePaymentConfirmationAvailable();
+</pre>
```
stephenmcgruer, line 653: Lets add something at the bottom like the following (but maybe with better wording than me :D):

```
NOTE: The use of the static {{PaymentRequest/isSecurePaymentConfirmationAvailable}} method is recommended for
      SPC feature detection, instead of calling {{PaymentRequest/canMakePayment}} on an already-constructed
      PaymentRequest object.
```

stephenmcgruer (approved):
LGTM with one comment and one optional comment. Thanks Nick!

spec.bs
nickburris: Done, that wording SGTM!

stephenmcgruer (approved):

stephenmcgruer:
@ianbjacobs - PTAL as well. Let us know also if you want this to wait for a WPWG meeting to let anyone chime in, or if you're happy to see it merged after your review.

spec.bs
```diff
@@ -622,6 +624,37 @@ members:
...
+
+<xmp class="idl">
+partial interface PaymentRequest {
+    static Promise<boolean> isSecurePaymentConfirmationAvailable();
+};
+</xmp>
+<dl dfn-type="attribute" dfn-for="PaymentRequest">
+    :   {{PaymentRequest/isSecurePaymentConfirmationAvailable()}}
+    ::  Upon invocation, a promise is returned that resolves with a value of
+        `true` if the user agent believes that the Secure Payment Confirmation
```
ianbjacobs, line 640: Is there a case where the UA would believe it is supported but it is not supported? If not, can we say "if the Secure Payment Confirmation feature is available" ?

stephenmcgruer: Removing 'believe' SGTM. If we look at ConditionalUI:

https://w3c.github.io/webauthn/#dom-publickeycredential-isconditionalmediationavailable

```
[PublicKeyCredential](https://w3c.github.io/webauthn/#publickeycredential) overrides this method to indicate availability for [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) mediation. [WebAuthn Relying Parties](https://w3c.github.io/webauthn/#webauthn-relying-party) SHOULD verify availability before attempting to set options.[mediation](https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-mediation) to [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional).

Upon invocation, a promise is returned that resolves with a value of true if [conditional](https://w3c.github.io/webappsec-credential-management/#dom-credentialmediationrequirement-conditional) [user mediation](https://w3c.github.io/webappsec-credential-management/#user-mediated) is available, or false otherwise.

This method has no arguments and returns a promise to a Boolean value.
```

So they just say 'is available'.

ianbjacobs (approved):

ianbjacobs:
Hi @nickburris,

I've approved the request, but this should not be part of the document that we advance to Candidate Recommendation. I do expect us to fold in other pull requests that will be part of the CR (namely, #230, #234, #236, and #238). Should we block merging this for now?

stephenmcgruer:
> Should we block merging this for now?

Yep; let's just leave this PR sitting until the first version of SPC gets to CR (assuming no significant bumps along the way). On the Chrome side, an open approved PR is sufficient for our launch processes :)

RByers:
[Chromium API owner hat on] Do y'all have an idea of how long you expect SPC to take to get to CR? While we can ship based on a PR if necessary, it's not ideal as it can contribute to confusion about what's ready to be implemented in other engines. If you expect it to take more than a month or so, could you consider creating a separate 'v2' branch or something which we can point to as the landed spec for this API?

ianbjacobs:
Hi @RByers,
We are poised to request to advance to CR. I am hopeful for a mid-May publication.

ianbjacobs:
@RByers, yes, very open to a versioning / branch discussion.

ianbjacobs:
@stephenmcgruer,
I'm ok to land the pull request, but would prefer to do so after the publication of the CR (ideally, therefore, in a few days). Would it be ok to wait for that? If you want to delegate landing this to me (for just after the publication), I'm happy to merge it.

stephenmcgruer:
Sure, sounds good to me - if you can take the task to land this once the CR is published, SGTM. Thanks! :)

PR #262, opened 2023-10-19
Update Github changelog link

rockymeza:
It looks like the CR branch is no longer being updated, so I thought this would be a better link.


ianbjacobs (approved):

rockymeza:
I have made a w3c account now. I think this is ready now

ianbjacobs:
Hi @rockymeza,

My recollection is that we are using the CR branch for the most recent Candidate Recommendation Snapshot, which is the last time we had confirmed sign-off from various W3C horizontal review groups. We have been publishing CR "drafts" since then, but not yet gotten new sign-off over the changes since the snapshot. I believe the current link is still relevant. The branch is not active, but it's not dead. When we next seek horizontal reviews (as part of advancing on the rec track), we will seek review based on the snapshot, and so the current diff link is still correct.

I'm going to close this issue; thanks for reaching out!

Ian

PR #272, opened 2024-08-13
Clarify that RP can use its own credentials even if extension not specified

ianbjacobs:
Make clear that an RP should be able to use its own credentials even if payment extension not specified.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -797,10 +797,15 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
 1. For each |id| in |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"]:
 
     1. Run the [=steps to silently determine if a credential is available for
-        the current device=] and the [=steps to silently determine if a
+        the current device=], passing in
+        |data|["{{SecurePaymentConfirmationRequest/rpId}}"] and |id|.
+	If the result is `false`, remove |id| from
+        |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"].
+    2.  If the |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is
+        not in the current origin, run the [=steps to silently determine if a
         credential is SPC-enabled=], passing in
         |data|["{{SecurePaymentConfirmationRequest/rpId}}"] and |id|. If the
-        result of either of these is `false`, remove |id| from
+        result is `false`, remove |id| from
         |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"].
 
 1. If |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"] is now empty,


scope.md
```diff
@@ -112,7 +112,7 @@ Note: This use case intends to capture the "in-transaction registration" use cas
 
 #### EMV&reg; Secure Remote Commerce (SRC) System as Relying Party
 
-* Alice checkouts on a merchant web site with SRC, which triggers the SRC Digital Card Facilitator (DCF) to be displayed. The SRC DCF asks whether she wants to use biometric authentication to streamline payment. She agrees and SRC DCF redirects her to her bank where she goes through an ID&V process with her bank for the credit card she wishes to use.
+* Alice checkouts on a merchant web site with SRC, which triggers the SRC Digital Card Facilitator (DCF) to be displayed. The SRC DCF asks whether she wants to use biometric authentication to streamline payment. She agrees and SRC DCF redirects her to her bank where she goes through an Identification and Verification process with her bank for the credit card she wishes to use.
```
stephenmcgruer, line 115: Do we need to spell out ID&V every time? Perhaps it could just go into https://w3c.github.io/secure-payment-confirmation/#sctn-terminology if needed (or be defined only the first time its used).

spec.bs
```diff
@@ -798,10 +798,15 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
 1. For each |id| in |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"]:
 
     1. Run the [=steps to silently determine if a credential is available for
-        the current device=] and the [=steps to silently determine if a
+        the current device=], passing in
+        |data|["{{SecurePaymentConfirmationRequest/rpId}}"] and |id|.
+        If the result is `false`, remove |id| from
+        |data|["{{SecurePaymentConfirmationRequest/credentialIds}}"].
+    1.  If the |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is
+        not in the current origin, run the [=steps to silently determine if a
```
stephenmcgruer, line 806: Yeah, so we don't just say things like "is not in the current origin" in specs 🤣 . We try to make them a bit more like code. I think this should be:

At the top of this method, change to:

```
The [=steps to validate payment method data=] for this payment method, for an
input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |data|, are:
```
(adding 'request', to get it in scope for looking up. This is a bit hand-wavy as Payment Request doesn't currently pass it in to the 'Steps to check if a payment can be made', but we can fix that later, not a big deal I think)

Then:

```
    1.  If the |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is
        not the [=origin=] of the [=relevant settings object=] of |request|,
        run the [=steps to silently determine if a
```

stephenmcgruer: Definitely not :). What we are trying to do here is determine if the origin that is **calling SPC** is the RP's origin, and thus should be able to skip the check for the thirdPartyPayment bit.

For example, consider the following nested case:

```
<-- main frame is merchant.example -->
  <iframe src="psp.example/pay">
    <iframe src="issuer.example/3ds">
    </iframe>
  </iframe>
```

For this setup, if `issuer.example` were to call SPC then origin would be `issuer.example` while the top-level origin would be `merchant.example`. But for this, we care about whether the caller themselves (`issuer.example`) is the owner of the passed-in credentials.

Does that make sense?

ianbjacobs: Yep. I was trying to solve a spec production error. I made this change and it went through! :) Clearly that was the wrong thing to do. However the error comes from using [=origin=]

stephenmcgruer: So the problem is that [=origin=] is ambiguous. Check out https://github.com/w3c/secure-payment-confirmation/actions/runs/11894955036/job/33143378569 , it does explain how to fix it:

image

So up in our [link-defaults block]](https://github.com/w3c/secure-payment-confirmation/blob/main/spec.bs#L88) we need to add an entry. I think one of the HTML ones is probably the most correct. Let's go with:

```
spec:html; type:dfn; for:/; text:origin
```

?

stephenmcgruer: Urgh, the HTML spec makes me sad (see the *new* errors :().

Sorry for the churn, what if we try the *other* suggested HTML default from bikeshed (the environment settings object one above)

spec.bs
```diff
@@ -206,7 +207,7 @@ This limitation motivates the following Secure Payment Confirmation behavior:
 
 1. SPC supports cross-origin registration from an iframe in a third-party
     context. For instance, this registration might take place following some
-    other identity and verification (<abbr>ID&amp;V</abbr>) flow (e.g., SMS OTP).
+    other identity and verification (<abbr>ID and V</abbr>) flow (e.g., SMS OTP).
```
stephenmcgruer, line 210: Nit; did this need changed or could it be left as-is?

stephenmcgruer (approved):
LGTM with one nit. Thank you!

spec.bs
ianbjacobs: I used ID&amp;V in the scope doc. I have no objection to leaving as-is here.

PR #280, opened 2025-02-07
[Spec] Separate 'SPC credential' from 'third-party enabled SPC credential'

stephenmcgruer:
This is a very minor change that is close to editorial, though it technically fixes a spec bug in "4.1.9 Steps to check if a payment can be made". Previously step 3.2 of that section only checked for an SPC Credential, not a third-party enabled SPC credential.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/280.html" title="Last updated on Feb 7, 2025, 6:17 PM UTC (9d6aa9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/280/0e2e5c5...9d6aa9c.html" title="Last updated on Feb 7, 2025, 6:17 PM UTC (9d6aa9c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -437,9 +437,7 @@ try {
 
 : <dfn export>SPC Credential</dfn>
 :: A [=public key credential|WebAuthn credential=] that can be used for the
-    behaviors defined in this specification. When an SPC Credential is to be
-    used by a party other than the [=Relying Party=], the [=Relying Party=] must
-    explicitly opt in by declaring the credential for SPC at creation time.
+    behaviors defined in this specification.
 
     This specification does not intend to limit how SPC credentials may (or may
     not) be used by a [=Relying Party=] for other authentication flows (e.g.,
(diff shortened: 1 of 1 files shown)


rsolomakhin (approved):

ianbjacobs:
@stephenmcgruer, seems good. Reading the algorithm it still looks like RP usage of their own credential is possible (in either 1p or 3p context). Can you confirm just so I'm sure? :)

ianbjacobs (approved):

stephenmcgruer:
> @stephenmcgruer, seems good. Reading the algorithm it still looks like RP usage of their own credential is possible (in either 1p or 3p context). Can you confirm just so I'm sure? :)

Correct. As long as the `steps to silently determine if a credential is available for the current device` return true for a given credential, then the RP of that credential can use it both in a 1p https://rp.example main-frame, and in a 3p https://rp.example iframe on another site (e.g., https://merchant.example). The `steps to silently determine if an SPC Credential is third-party enabled` only apply to if https://merchant.example tries to use a credential from another RP.

Issue #284, opened 2025-02-24
Should/could isSecurePaymentConfirmationAvailable return an enum rather than a boolean?


stephenmcgruer:
Currently, [isSecurePaymentConfirmationAvailable](https://w3c.github.io/secure-payment-confirmation/#sctn-secure-payment-confirmation-available-api) returns a `Promise<boolean>`, indicating "if the Secure Payment Confirmation feature is available, or false otherwise".

We have been recently implementing this feature in Chrome, and during implementation found that it may be useful to know more than just `true` or `false`, but also some details on *why* SPC isn't available. For us during implementation this would have helped us debug code bugs, but for web developers we could imagine being able to understand + explain to users why SPC isn't available to them on the current device, or even just to record statistics across their users.

This might also help address reasonable concerns that were [raised on the issue](https://github.com/w3c/secure-payment-confirmation/issues/81#issuecomment-1527105768) relating to being clearer in the spec as to how this API should operate.

At the same time, we should be aware of potential privacy risks here, so will need to evaluate the potential for fingerprinting.

stephenmcgruer:
Using the current Chrome implementation as a guide, when isSecurePaymentConfirmationAvailable is called, we check:

1. If the [SecurePaymentConfirmation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=3854;drc=d7f594ac26932bad60ba3e8d1754d36e6b511c22) feature is not enabled, return false ([payment_request.cc:836](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/payments/payment_request.cc;l=836;drc=05401be858fce13f572977457ff78cea16418569)
    - This feature is technically controllable by users via a command-line flag, but for most users it corresponds to what device they are on. It is enabled for Mac, Windows, and Android builds of Chromium, and disabled otherwise ([content_features.cc:991](https://source.chromium.org/chromium/chromium/src/+/main:content/public/common/content_features.cc;l=991;drc=9cd9facf47fe8d99938a1d9fea8798a9fc1f9b64))

2. If the `"payment"` Permission Policy is not enabled on the current frame, return false ([payment_request.cc:842](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/payments/payment_request.cc;l=842;drc=05401be858fce13f572977457ff78cea16418569)
 
[...]

PR #285, opened 2025-03-10
[Spec] Add output enum for isSecurePaymentConfirmationAvailable

stephenmcgruer:
Fixes https://github.com/w3c/secure-payment-confirmation/issues/284

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -73,6 +73,7 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: authentication extension; url: authentication-extension
         text: extension identifier; url: extension-identifier
         text: user member; url: dom-publickeycredentialcreationoptions-user
+        text: user-verifying platform authenticator; url: user-verifying-platform-authenticator
 
 spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
@@ -365,7 +366,7 @@ presumes access to await/async, for easier to read promise handling.
 const spcAvailable =
   PaymentRequest &&
   PaymentRequest.isSecurePaymentConfirmationAvailable &&
-  await PaymentRequest.isSecurePaymentConfirmationAvailable();
+  (await PaymentRequest.isSecurePaymentConfirmationAvailable()) === 'available';
 if (!spcAvailable) {
   /* Browser does not support SPC; merchant should fallback to traditional flows. */
 }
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
@rsolomakhin @ianbjacobs - PTAL

cc @pejic

ianbjacobs (commented):
@stephenmcgruer,

Thanks for this pull request. Some questions:

* What is the relationship between unavailable-no-user-verifying-platform-authenticator and isUserVerifyingPlatformAuthenticatorAvailable() ? Could the latter be called outside of SPC (and thus moot the need for the SPC value)?

* I think the text should explain why there is an enum. At a high level this is meant to help developers without leaking additional information. But what's not clear from the text is the developer needs or use cases driving the move to the enum.

stephenmcgruer:
Thanks Ian! I'll address your latter bullet shortly, but for the first:

> What is the relationship between unavailable-no-user-verifying-platform-authenticator and isUserVerifyingPlatformAuthenticatorAvailable() ? Could the latter be called outside of SPC (and thus moot the need for the SPC value)?

They are identical, but it doesn't feel like it makes sense to me to have spcAvailability == 'available' but then you also have to check `isUserVerifyingPlatformAuthenticatorAvailable()` and if that returns false SPC will still not work?

Getting into the weeds, I could even see it making more sense to return a set of results, since technically you could have SPC not working for multiple reasons. But at some point we start getting into silly territory, when we started with a simple true/false boolean!

ianbjacobs:
> They are identical, but it doesn't feel like it makes sense to me to have spcAvailability == 'available' but then you also have to check isUserVerifyingPlatformAuthenticatorAvailable() and if that returns false SPC will still not work?

I suggest simply calling out the relationship to avoid questions.

stephenmcgruer:
Ack Ian, sounds good and I'll address that and re-ping.

cc @domenic also, just because I'm curious as to whether you have any "best practice" thoughts on this sort of API. A few things I'm only musing now (whoops; we live and we learn 🤦 ): (1) _should_ it be a set of reasons?, and (2) if not, should the spec be explicit about the ordering of priority (e.g., if feature is not enabled return that, else if permission policy not set return that, else if ... ).

domenic:
Easy answers first...

> 2) if not, should the spec be explicit about the ordering of priority (e.g., if feature is not enabled return that, else if permission policy not set return that, else if ... )

The spec should indeed be explicit about the ordering. Ideally, this would even be true if you did a set of reasons! Since it's possible to imagine web developers doing something like `if (reasonSet[0] === "some-reason") { ... }` and that code working in only one browser.

(An easy way to order sets is by sorting them, though.)


> (3) should this be called `getSecurePaymentConfirmationAvailability()` instead of just `securePaymentConfirmationAvailability`...

I like omitting the `get` prefix, personally.

---

Harder question:

> (1) _should_ it be a set of reasons?

This seems like it depends on your use case. What code do you expect developers to write? Is this for reporting to the server, so that they can aggregate all the failures and get insight into statistical patterns? If so, a set is often useful. An example is bfcache blocking reasons.

Or is this for runtime determination of what UI to show? If so, it's a bit harder to imagine how listing all the reasons would help. (But maybe it would; I'm not the domain expert!)

stephenmcgruer:
Thanks Dominic! I chatted this over with our privacy reviewer, and given that there's a *slight* increase in fingerprinting potential by allowing multiple reasons to be listed, plus a solid increase in implementation complexity and no clear developer need yet, I opted for just making the order specific but leaving it as a single outcome for now. Thank you for your input!

stephenmcgruer:
@ianbjacobs - I have addressed your comments, please take another look :)

spec.bs
```diff
@@ -656,33 +657,103 @@ The [=payment method additional data type=] for this payment method is
...
+<dl dfn-type="enum-value" dfn-for="SecurePaymentConfirmationAvailability">
+    :   <dfn>available</dfn>
+    ::  Indicates that the user agent believes that the Secure Payment
+        Confirmation API is available in the calling frame.
+
+        Note: This result does not indicate whether or not any particular [=SPC
+              credential=] is or will be available.
+
+    :   <dfn>unavailable-unknown-reason</dfn>
+    ::  Indicates that the Secure Payment Confirmation API is not avaiable in
```
ianbjacobs, line 691:
```suggestion
    ::  Indicates that the Secure Payment Confirmation API is not available in
```

spec.bs
```diff
@@ -1548,6 +1619,40 @@ they are strong, cross-site identifiers. However in order to obtain them from
...
+
+- {{SecurePaymentConfirmationAvailability/unavailable-feature-not-enabled}}:
+    some risk of fingerprinting, depending on under what circumstances the user
+    agent considers Secure Payment Confirmation to be available or not. User
+    agents are encouraged to make Secure Payment Confirmation available to all
+    users (if implementing the specification), or at least to significantly
+    sized groups such that no (additional) fingerprinting is possible. For
+    example, a user agent may ship Secure Payment Confirmation to all users on
+    a given OS but not others - this then reduces the fingerprinting risk to
+    no more than the user agent string already reveals.
```
ianbjacobs, line 1638:
```suggestion
    no more than what the user agent string already reveals.
```

ianbjacobs (approved):
See two typo fixes, otherwise thanks!!

Issue #271, opened 2024-05-23
Proposal: WebAuthn-agnostic browser binding for Secure Payment Confirmation


stephenmcgruer:
# Proposal: WebAuthn-agnostic <s>device</s> browser binding for Secure Payment Confirmation

This is a high-level proposal for adding device-binding-like capabilities, in the form of **browser bound keys** (BBKs), to Secure Payment Confirmation without relying on WebAuthn (at either the client or authenticator level). At time of writing it is still exploratory and should not be taken as a concrete plan until the WG has discussed and decides it is the appropriate direction to move in. The proposal is also missing detail that will need to be filled in before it could be spec'd / implemented.

## Background / Motivation

Secure Payment Confirmation builds on top of WebAuthn to make it more useful for the payments industry. Today, that comes via (at least) three added features on-top of WebAuthn:

1. Providing a consistent and clear browser-hosted transaction dialog for the user to review their transaction details and agree to the amount, payee, etc.
1. Including transaction details in the WebAuthn cryptogram (as part of ClientDataJSON), with a goal of helping with payment requirements such as Dynamic Linking.
1. Supporting cross-origin authentication flows, such that a merchant or PSP can initiate an SPC authentication in their domain with a relying party of (for example) the issuing bank, without having to iframe or redirect to the issuing bank.

As of late 2023/early 2024 however, we are hearing strongly from our payments partners that a significant required feature is missing - device binding. As WebAuthn passkeys can now be synced, it can be argued that they no longer meet strict 2FA requirements (being no longer a signal of device possession), and so SPC (like WebAuthn) is reduced to a single factor (biometric or possession, depending on the authentication method used).

WebAuthn/FIDO have been exploring device binding for years, and have a proposed solution in the form of DPK/[SPK](https://w3c.github.io/webauthn/#sctn-supplemental-public-keys-extension). However, progress in making DPK/SPK widely available/deployed appears slow. As of May 2024 no platform authenticator ships DPK/SPK, and support is not (as far as the author knows) expected across the major platform authenticators this year - not is there any commitment or expectation from any platform authenticator that they will ship DPK/SPK.

## Proposal

Given the timescales and uncertainty involved with WebAuthn's SPK, develop an SPC-specific <s>device</s> browser binding solution that can be experimented with and potentially launched during 2024. 

As a rough outline, during SPC authentication in the case where the passed-in WebAuthn credential is available and the user has accepted the SPC transaction dialog, the user agent would:

[...]

PR #286, opened 2025-03-17
Add browser binding to the secure payment confirmation spec

pejic:
See #271

***
<a href="https://pr-preview.s3.amazonaws.com/pejic/secure-payment-confirmation/pull/286.html" title="Last updated on Jun 5, 2025, 5:16 PM UTC (c527dbf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/286/c7dd82e...pejic:c527dbf.html" title="Last updated on Jun 5, 2025, 5:16 PM UTC (c527dbf)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -405,7 +405,7 @@ const request = new PaymentRequest([{
     ],
 
     // Caller’s requested localized experience
-    locale: ["en"], 
+    locale: ["en"],
 
     timeout: 360000,  // 6 minutes
   }], {
--- a/spec.bs
+++ b/spec.bs
@@ -44,6 +44,10 @@ spec: i18n-glossary; urlPrefix: https://www.w3.org/TR/i18n-glossary/
     type: dfn
         text: locale; url: dfn-locale
 
+spec: webcrypto; urlPrefix: https://www.w3.org/TR/webcrypto-2/
+    type: idl
+        text: getRandomValues
+
 spec: payment-request; urlPrefix: https://w3c.github.io/payment-request/
     type: dfn
         text: payment method; url: dfn-payment-method
@@ -76,6 +80,11 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: extension identifier; url: extension-identifier
         text: user member; url: dom-publickeycredentialcreationoptions-user
         text: user-verifying platform authenticator; url: user-verifying-platform-authenticator
+        text: create credential async loop; url: CreateCred-async-loop
+        text: credentialPublicKey; url: authdata-attestedcredentialdata-credentialpublickey
+        text: credentialIdResult; url: assertioncreationdata-credentialidresult
+        text: base64url encoding; url: base64url-encoding
+        text: COSEAlgorithmIdentifier
 
 spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
@@ -93,6 +102,8 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
 spec:fetch; type:dfn; for:/; text:request;
 spec:i18n-glossary; type:dfn; text:bidi isolation
 spec:i18n-glossary; type:dfn; text:language priority list
+spec:infra; type:dfn; text:list
+spec:infra; type:dfn; text:user agent
 spec:url; type:dfn; text:valid domain;
 spec:html; type:dfn; for:environment settings object; text:origin
 </pre>
@@ -107,6 +118,12 @@ spec:html; type:dfn; for:environment settings object; text:origin
     "IANA-WebAuthn-Registries": {
         "title": "Registries for Web Authentication (WebAuthn)",
         "href": "https://www.rfc-editor.org/rfc/rfc8809.html"
+    },
+
+    "IANA-COSE-ALGS-REG": {
+        "href": "https://www.iana.org/assignments/cose/cose.xhtml#algorithms",
+        "title": "IANA CBOR Object Signing and Encryption (COSE) Algorithms Registry",
+        "publisher": "IANA"
     }
 }
 </pre>
@@ -143,8 +160,8 @@ opt-in to allowing this behavior during credential creation.
 
 Functionally, this specification defines a new [=payment method=] for the
 {{PaymentRequest}} API, and adds a [=WebAuthn Extension=] to extend
-[[webauthn-3]] with payment-specific datastructures and to relax assumptions to
-allow the API to be called in payment contexts.
+[[webauthn-3]] with payment-specific datastructures, device binding and to
+relax assumptions to allow the API to be called in payment contexts.
 
 ## Use Cases ## {#sctn-use-cases}
 
@@ -220,6 +237,20 @@ payment service providers are likely to build them on behalf of merchants.
 
 NOTE: Relying Parties that wish to provide the authentication user experience
 may still do so using SPC from an iframe.
+
+### Cryptographic evidence of device binding ### {#sctn-use-case-device-binding}
+
+In the payments industry a signal of device possession plays an important role
+as a second factor. WebAuthn allows for synced passkeys where one credential is
+available on multiple devices ([[webauthn-3#sctn-usecase-consumer-mdc]]).
+Although syncing improves the user experience for login use cases, concerns
+have been raised that synced passkeys alone do not satisfy device possession
+requirements in some regulatory environments.
+
+These concerns motivate the inclusion of auxiliary [=key pair|public private
+key pairs=] created by the user agent where the private key only resides on
+(and is used on) one device. Such a key and its use in SPC is referred to as a
+[=browser bound key=].
 
 ## Sample API Usage Scenarios ## {#sctn-sample-scenarios}
 
@@ -315,6 +346,19 @@ const publicKey = {
   extensions: {
     "payment": {
       isPayment: true,
+      // An optional list of allowed algorithms. When not present or empty, the
+      // pubKeyCredparams are used defaulting to ES256 and RS256. In this
+      // example ES256 and RS256 are allowed and RS256 is preferred.
+      browserBoundPubKeyCredParams: [
+        {
+          type: "public-key",
+          alg: -257 // "RS256"
+        },
+        {
+          type: "public-key",
+          alg: -7 // "ES256"
+        }
+      ]
     }
   }
 };
@@ -351,13 +395,16 @@ Payment Confirmation.
 
 1. The user agrees to the payment-specific information displayed in the SPC UX,
     and performs a subsequent WebAuthn authentication ceremony. The signed
-    cryptogram is returned to the merchant.
+    cryptogram is returned to the merchant (including browser bound key outputs
+    in {{AuthenticationExtensionsPaymentOutputs}}).
 
 1. The merchant communicates the signed cryptogram to the issuing bank
     out-of-band. The issuing bank verifies the cryptogram, and knows that the
     user is valid, what payment-specific information has been displayed, and
     that the user has consented to the transaction. The issuing bank authorizes
     the transaction and the merchant finishes the checkout process for the user.
+    The issuing bank stores the [=browser bound key=]'s public key,
+    {{CollectedClientAdditionalPaymentData/browserBoundPublicKey}}.
 
 The sample code for authenticating the user follows. Note that the example code
 presumes access to await/async, for easier to read promise handling.
@@ -408,6 +455,22 @@ const request = new PaymentRequest([{
     locale: ["en"],
 
     timeout: 360000,  // 6 minutes
+
+    // An optional list of allowed algorithms defaulting to ES256 and RS256.
+    // In this example ES256 and RS256 are allowed and ES256 is preferred.
+    // Browser bound keys are not created when already present, so this
+    // list is only used when the browser bound key does need to be
+    // created.
+    browserBoundPubKeyCredParams: [
+      {
+        type: "public-key",
+        alg: -7 // "ES256"
+      },
+      {
+        type: "public-key",
+        alg: -257 // "RS256"
+      }
+    ]
   }], {
     total: {
       label: "Total",
@@ -482,6 +545,30 @@ try {
     NOTE: This property is very similar to that which is required for the
     [[webauthn-conditional-ui inline]]. It is likely that both it and SPC could
     be supported by the same underlying API.
+
+: <dfn>Browser bound key</dfn>
+:: A public-private [=key pair=] that signs over the transaction details in
+    addition to the WebAuthn credential and is tied to a single device by the
+    user agent.
+
+: <dfn>Key Pair</dfn>
+:: A key pair is a [=struct=] consisting of <dfn>public key</dfn> and
+    <dfn>private key</dfn>, implementation defined types.
+
+    Note: [=Public key=] and [=private key=] are algorithm specific parameters
+        to the cryptographic algorithms referenced by the IANA COSE Algorithms
+        registry [[!IANA-COSE-ALGS-REG]].
+
+    Note: The [=public key=] portion is returned during registration in
+        {{CollectedClientAdditionalPaymentRegistrationData}}.{{CollectedClientAdditionalPaymentRegistrationData/browserBoundPublicKey}}
+        and during payment assertion in
+        {{CollectedClientAdditionalPaymentData}}.{{CollectedClientAdditionalPaymentData/browserBoundPublicKey}}.
+
+    Note: The [=private key=] is used to generate cryptographic signatures
+        included in
+        {{BrowserBoundSignature}}.{{BrowserBoundSignature/signature}}. The user
+        agent does not export the private key and may store the private key in
+        a secure element of the device.
 
 # Registration # {#sctn-registration}
 
@@ -594,6 +681,7 @@ NOTE: This allows the user agent to not require user activation, for
         USVString payeeOrigin;
         sequence<PaymentEntityLogo> paymentEntitiesLogos;
         AuthenticationExtensionsClientInputs extensions;
+        sequence<PublicKeyCredentialParameters> browserBoundPubKeyCredParams;
         sequence<USVString> locale;
         boolean showOptOut;
     };
@@ -651,6 +739,13 @@ members:
         credential(s). The caller does not need to specify the
         [[#sctn-payment-extension-registration| payment extension]]; it is added
         automatically.
+
+    :  <dfn>browserBoundPubKeyCredParams</dfn> member
+    :: The list of allowed types of credential restricting the types of
+        cryptographic algorithms used for the browser bound key.
+
+        Note: This member will be used only if a browser bound key will need to
+              be created.
 
     :  <dfn>locale</dfn> member
     :: An optional list of well-formed [[BCP47]] language tags, in descending
@@ -1106,6 +1201,8 @@ Note: These steps are only run if the user accepts the
     :: |request|.[=payment request details|[[details]]=]["{{PaymentDetailsInit/total}}"]
     : {{AuthenticationExtensionsPaymentInputs/instrument}}
     :: |data|["{{SecurePaymentConfirmationRequest/instrument}}"]
+    : {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+    :: |data|["{{SecurePaymentConfirmationRequest/browserBoundPubKeyCredParams}}"]
 
 1. Let |extensions| be a new {{AuthenticationExtensionsClientInputs}} dictionary
     whose {{AuthenticationExtensionsClientInputs/payment}} member is set to
@@ -1197,6 +1294,7 @@ Note: Previously, the `payment` extension allowed for the creation of
 
     dictionary AuthenticationExtensionsPaymentInputs {
       boolean isPayment;
+      sequence<PublicKeyCredentialParameters> browserBoundPubKeyCredParams;
 
       // Only used for authentication.
       USVString rpId;
@@ -1234,6 +1332,13 @@ Note: Previously, the `payment` extension allowed for the creation of
       :  <dfn>instrument</dfn> member
       :: The instrument details that were displayed to the user. Only used at authentication time; not registration.
 
+      :  <dfn>browserBoundPubKeyCredParams</dfn> member
+      :: The list of allowed types of credential restricting the types of
+         cryptographic algorithms used for the browser bound key.
+
+         Note: When this member is not present, it defaults to
+         {{PublicKeyCredentialCreationOptions}}.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
     </dl>
 
 :  <dfn export>Client extension processing ([=registration extension|registration=])</dfn>
@@ -1257,7 +1362,60 @@ Note: Previously, the `payment` extension allowed for the creation of
               enrollment.https.html
             </wpt>
 
-:  <dfn export>Client extension processing ([=authentication extension|authentication=])</dfn>
+    1. Before step 13 (creating a {{CollectedClientData}}).
+
+        1. Let |bbk_allowed_algorithms| be {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}.
+
+        1. If
+            {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+            is empty, let |bbk_allowed_algorithms| be
+            {{PublicKeyCredentialCreationOptions}}.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+        1. Let |bbk_and_algorithm|, |bbk_id| be the result of [=creating a key
+            pair=] using |bbk_allowed_algorithms|.
+
+        1. If the result of [=creating a key pair=] is null, skip the steps
+            below relating to |bbk_and_algorithm| and |bbk_public_key|.
+
+        1. Let |bbk_public_key| be the result of [=getting a browser bound
+            public key=] using |bbk_and_algorithm|.
+
+    1. In step 13, instead of creating the {{CollectedClientData}} create a
+        {{CollectedClientPaymentData}} whose fields are:
+
+        1. {{CollectedClientPaymentData/payment}} set to a
+            {{CollectedClientAdditionalPaymentRegistrationData}} with
+
+            : {{CollectedClientAdditionalPaymentRegistrationData/browserBoundPublicKey}}
+            :: |bbk_public_key| - The COSE_Key encoded [=public key=].
+
+        1. All other fields set as per the original step 13.
+
+    1. In [=create credential async loop|step 22=], case "any authenticator
+        indicates success", step 3, just before returning <var
+        ignore>pubKeyCred</var>:
+
+        1. [=Bind a key pair=] using |bbk_id| and <var
+            ignore>pubKeyCred</var>.{{PublicKeyCredential/[[identifier]]}}. If
+            the result is failure, then return {{UnknownError}}.
+
+        1. Let |payment_outputs| be {{AuthenticationExtensionsPaymentOutputs}} with:
+
+            : {{AuthenticationExtensionsPaymentOutputs/browserBoundSignature}}
+            :: A {{BrowserBoundSignature}} with fields:
+
+                : {{BrowserBoundSignature/signature}}
+                :: Result of [=generating a browser bound signature=] using
+                    |bbk_and_algorithm| and |clientDataJson| from
+                    [[webauthn-3#sctn-createCredential|create credential]]
+                    step&nbsp;14.
+
+        1. Set
+            [[#sctn-payment-extension-registration|"payment"]] →
+            |payment_outputs| on the
+            {{PublicKeyCredential/[[clientExtensionsResults]]}}.
+
+: <dfn export>Client extension processing ([=authentication extension|authentication=])</dfn>
 :: When [[webauthn-3#sctn-getAssertion|making an assertion]] with a
     {{AuthenticationExtensionsPaymentInputs}} |extension_inputs|:
 
@@ -1283,7 +1441,20 @@ Note: Previously, the `payment` extension allowed for the creation of
             Note: This enables cross-domain authentication ceremonies; see
                   [[#sctn-use-case-merchant-authentication]].
 
-        1. In step 9, instead of creating a {{CollectedClientData}}, instead
+        1. Before step 10, before creating a {{CollectedClientData}}.
+
+            1. Let |allowed_algorithms| be the
+                {{SecurePaymentConfirmationRequest}}.{{SecurePaymentConfirmationRequest/browserBoundPubKeyCredParams}}.
+
+            1. Let |bbk_and_algorithm| be the result of [=get or create a browser bound
+                key|getting or creating a browser bound key=] using
+                |credential_id| and |allowed_algorithms|.
+
+            1. If |bbk_and_algorithm| is null, do not include the
+                {{CollectedClientAdditionalPaymentData/browserBoundPublicKey}}
+                in subsequent steps.
+
+        1. In step 10, instead of creating a {{CollectedClientData}}, instead
             create a {{CollectedClientPaymentData}} with:
 
             1. {{CollectedClientData/type}} set to "`payment.get`"
@@ -1307,15 +1478,64 @@ Note: Previously, the `payment` extension allowed for the creation of
                 :: |extension_inputs|["{{AuthenticationExtensionsPaymentInputs/total}}"]
                 : {{CollectedClientAdditionalPaymentData/instrument}}
                 :: |extension_inputs|["{{AuthenticationExtensionsPaymentInputs/instrument}}"]
+                : {{CollectedClientAdditionalPaymentData/browserBoundPublicKey}}
+                :: If a bbk was created, the result of [=getting a browser
+                    bound public key=].
 
-            1. All other fields set as per the original step 9.
+            1. All other fields set as per the original step 10.
 
             <wpt>
               authentication-accepted.https.html
             </wpt>
 
+        1. During "If any authenticator indicates success"; in step&nbsp;2 when
+            setting the clientExtensionResults.
+
+            1. Let |payment_outputs| be
+                {{AuthenticationExtensionsPaymentOutputs}} with:
+
+                : {{AuthenticationExtensionsPaymentOutputs/browserBoundSignature}}
+                :: A {{BrowserBoundSignature}} with fields:
+
+                    : {{BrowserBoundSignature/signature}}
+                    :: Result of [=generating a browser bound
+                        signature=] using |bbk_and_algorithm| and
+                        |clientDataJson| from
+                        [[webauthn-3#sctn-createCredential|create credential]]
+                        step&nbsp;14.
+
+            1. Set
+                [[#sctn-payment-extension-registration|"payment"]] →
+                |payment_outputs| on the
+                {{PublicKeyCredential/[[clientExtensionsResults]]}}.
+
 :  Client extension output
-:: None
+::
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      AuthenticationExtensionsPaymentOutputs payment;
+    };
+
+    dictionary AuthenticationExtensionsPaymentOutputs {
+      BrowserBoundSignature browserBoundSignature;
+    };
+
+    dictionary BrowserBoundSignature {
+      required ArrayBuffer signature;
+    };
+    </xmp>
+
+    <dl dfn-type="dict-member" dfn-for="AuthenticationExtensionsPaymentOutputs">
+      :  <dfn>browserBoundSignature</dfn> member
+      :: Outputs of the browser bound signature procedure.
+
+    </dl>
+
+    <dl dfn-type="dict-member" dfn-for="BrowserBoundSignature">
+      :  <dfn>signature</dfn> member
+      :: The output of the browser bound signing process.
+
+    </dl>
 
 :  Authenticator extension processing
 :: None
@@ -1324,7 +1544,7 @@ Note: Previously, the `payment` extension allowed for the creation of
 
 <xmp class="idl">
     dictionary CollectedClientPaymentData : CollectedClientData {
-        required CollectedClientAdditionalPaymentData payment;
+        required (CollectedClientAdditionalPaymentData or CollectedClientAdditionalPaymentRegistrationData) payment;
     };
 </xmp>
 
@@ -1347,6 +1567,7 @@ The {{CollectedClientPaymentData}} dictionary inherits from
         sequence<PaymentEntityLogo> paymentEntitiesLogos;
         required PaymentCurrencyAmount total;
         required PaymentCredentialInstrument instrument;
+        USVString browserBoundPublicKey;
     };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -218,6 +218,18 @@ payment service providers are likely to build them on behalf of merchants.
 NOTE: Relying Parties that wish to provide the authentication user experience
 may still do so using SPC from an iframe.
 
+### Cryptographic evidence of device binding ### {#sctn-use-case-device-binding}
```
stephenmcgruer, line 221: @ianbjacobs - would love your thoughts, on this PR in general but particularly on the non-normative sections where it speaks to the motivation + needs

spec.bs
```diff
@@ -218,6 +218,18 @@ payment service providers are likely to build them on behalf of merchants.
...
+
+In the payments industry a signal of device possession is used as a second
```
ianbjacobs, line 223:
```suggestion
In the payments industry a signal of device possession plays an important role as a second
```

spec.bs
```diff
@@ -468,6 +515,19 @@ try {
...
     be supported by the same underlying API.
 
+: <dfn>Browser bound key</dfn>
+:: A public-private key pair that signs over the transaction in addition to the
+    WebAuthn credential and is tied to a single device by the user agent.
+
+: <dfn>Platform Cryptographic APIs</dfn>
+:: Operating system APIs providing cryptographic algorithms and storage.
+    Particularly browser bound keys require APIs for generating key pairs,
+    retrieving key pairs, exporting public keys, and generating signatures.
```
ianbjacobs, line 525:
```suggestion
    For browser bound keys, the user agent uses these APIs to generate key pairs,
    retrieve key pairs, export public keys, and generate signatures.
```

spec.bs
```diff
@@ -468,6 +515,19 @@ try {
     [[webauthn-conditional-ui inline]]. It is likely that both it and SPC could
     be supported by the same underlying API.
 
+: <dfn>Browser bound key</dfn>
+:: A public-private key pair that signs over the transaction in addition to the
+    WebAuthn credential and is tied to a single device by the user agent.
+
+: <dfn>Platform Cryptographic APIs</dfn>
```
stephenmcgruer, line 522: This is never referenced in the text; is it valuable to have it defined here? Notably, I don't think WebAuthn defines something like this, although perhaps this is because they have the [Authenticator model](https://w3c.github.io/webauthn/#sctn-authenticator-model) instead (which may be useful as a template for some of this work!)

spec.bs
```diff
@@ -468,6 +515,19 @@ try {
...
+:: Operating system APIs providing cryptographic algorithms and storage.
+    Particularly browser bound keys require APIs for generating key pairs,
+    retrieving key pairs, exporting public keys, and generating signatures.
+
+: <dfn>Secure element</dfn>
```
stephenmcgruer, line 527: Similarly, I'm not sure if this is worth/needs to be defined. WebAuthn spec talks about + refers to "Secure Element", "TPM", etc quite freely without defining them (ctrl+f on https://w3c.github.io/webauthn)

spec.bs
```diff
@@ -626,6 +686,9 @@ members:
     :: Any [=WebAuthn extensions=] that should be used for the passed
         credential(s). The caller does not need to specify the
         [[#sctn-payment-extension-registration| payment extension]]; it is added
+        automatically. When setting a non-default
+        {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+        then {{AuthenticationExtensionsPaymentInputs/isPayment}} is set
         automatically.
 
     :  <dfn>locale</dfn> member
```
stephenmcgruer, line 694: (I dislike how GitHub doesn't let me tag comments to lines outside of the git diff, even if you expand the file...)

Do we need to do any validation of `browserBoundPubKeyCredParams` from the SecurePaymentConfirmationRequest dictionary during **4.1.8. Steps to validate payment method data** (line 749)? For example, what if the sequence is empty?

spec.bs
```diff
@@ -997,6 +1060,7 @@ Note: Previously, the `payment` extension allowed for the creation of
       USVString payeeOrigin;
       PaymentCurrencyAmount total;
       PaymentCredentialInstrument instrument;
+      PublicKeyCredentialParameters browserBoundPubKeyCredParams;
```
stephenmcgruer, line 1063: Should this be a `sequence<PublicKeyCredentialParameters>` ? Right now its a single object.

Also, it shouldn't be under `// Only used for authentication.`; either add a new section with another comment, or (preferred, I think?) put it above between isPayment and rpId.

spec.bs
```diff
@@ -1022,6 +1086,38 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+
+:  Client extension outputs
+::
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      AuthenticationExtensionsPaymentOutputs payment;
+    };
+
+    dictionary AuthenticationExtensionsPaymentOutputs {
+      AuthenticationExtensionPaymentBrowserBoundSignature browserBoundSignature;
```
stephenmcgruer, line 1103: Bikeshedding, but we don't need to specifically prefix `AuthenticationExtensionPayment` here, we can just call this `BrowserBoundSignature` at this point or similar.

spec.bs
```diff
@@ -1022,6 +1086,38 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+    };
+
+    dictionary AuthenticationExtensionPaymentBrowserBoundSignature {
+      required ArrayBuffer signatureOutput;
```
stephenmcgruer, line 1107: Bikeshedding, but WebAuthn just calls their signature arraybuffers `signature`, so I suggest we do the same (i.e. drop `Output`).

spec.bs
```diff
@@ -1022,6 +1086,38 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+    <xmp class="idl">
+    partial dictionary AuthenticationExtensionsClientOutputs {
+      AuthenticationExtensionsPaymentOutputs payment;
+    };
+
+    dictionary AuthenticationExtensionsPaymentOutputs {
+      AuthenticationExtensionPaymentBrowserBoundSignature browserBoundSignature;
+    };
+
+    dictionary AuthenticationExtensionPaymentBrowserBoundSignature {
```
stephenmcgruer, line 1106: Why is this nested dictionary needed? Is it likely that there will be other members in the future aside from the `signatureOutput` ?

spec.bs
```diff
@@ -1045,7 +1141,66 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+                :: Skip steps below relating to |bbk| and |bbk_public_key|.
+
+            </dl>
+
+        1. Let |bbk_public_key| be the result of
+            {{BrowserBoundKeyStore/GetBrowserBoundPublicKey}}(|bbk|).
+
+    1. In step 13, instead of creating the {{CollectedClientData}} create a {{CollectedClientPaymentData}} whose fields are:
+
+        1. {{CollectedClientPaymentData/payment}} set to a {{CollectedClientAdditionalPaymentData}} with
```
stephenmcgruer, line 1164: Ah, this is tricky. `CollectedClientAdditionalPaymentData` has a bunch of `required` members currently, and so you cannot construct one with just the `browserBoundPublicKey` field set. I don't have an answer for this immediately, we'll have to figure out what we want to do here (include the other members? make them optional? - technically backwards incompatible but only very minorly given this is basically an internal data structure. or define a different CollectedClientAdditionalPaymentData for registration time?)

spec.bs
```diff
@@ -1045,7 +1141,66 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+
+            : {{CollectedClientAdditionalPaymentData/browserBoundPublicKey}}
+            :: |bbk_public_key| - The COSE_Key encoded public key.
+
+    1. In step 22 (webnauthn-3#CreateCred-async-loop), when "If any
```
stephenmcgruer, line 1169: Could maybe just say "Just before creating a [PublicKeyCredential](https://w3c.github.io/webauthn/#publickeycredential), do [...]"

See also how WebAuthn has done this for https://w3c.github.io/webauthn/#sctn-appid-exclude-extension - albeit they added bookmarks into their spec to make their life easier, I think? (need to check out their bikeshed source)

spec.bs
```diff
@@ -1131,6 +1334,7 @@ The {{CollectedClientPaymentData}} dictionary inherits from
         USVString payeeOrigin;
         required PaymentCurrencyAmount total;
         required PaymentCredentialInstrument instrument;
+        USVString browserBoundPublicKey;
```
stephenmcgruer, line 1337: This needs defined in the list below on line 1345-1359

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
 {{CollectedClientAdditionalPaymentData}}, because the origin of the calling
 frame is already included in {{CollectedClientData}} of [[webauthn-3]].
 
+# Browser Bound Key Store - <dfn interface>BrowserBoundKeyStore</dfn> Internal Component # {#sctn-browser-bound-key-store}
+
+This component in the user agent manages platform-dependant cryptographic
+keypairs including their association to SPC credentials (i.e. passkeys). This
+section explains the |BrowserBoundKeyStore| and procedures for creating,
+binding, retrieving, and signing using the browser bound keypairs.
```
stephenmcgruer, line 1377: I think "keypair" is important enough that we need to define it in terminology above. We can borrow from WebAuthn who have "Credential Key Pair" (https://w3c.github.io/webauthn/#credential-key-pair); we want basically the same but without 'credential'.

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
...
+
+The browser bound key store contains |browser_bound_map|, a map from
```
stephenmcgruer, line 1379: We should not comment on where/how the user agent stores the map. See WebAuthn which has a very similar concept in its credentials map (https://w3c.github.io/webauthn/#authenticator-credentials-map)

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
...
+
+1. If the |allowed_algorithms| are an empty list, set this list to a
+    default list containing the following in order:
+    * {{PublicKeyCredentialParameters/type}} = {{PublicKeyCredentialType}} with {{PublicKeyCredentialParameters/alg}} = -7 ("ES256").
+    * {{PublicKeyCredentialParameters/type}} = {{PublicKeyCredentialType}} with {{PublicKeyCredentialParameters/alg}} = -257 ("RS256").
+
+1. Remove any entries from |allowed_algorithms| that are not of type
+    {{PublicKeyCredential}}.
+
+    Issue: <span style="background-color: red">TODO</span>: Should we expand
```
stephenmcgruer, line 1401: I think we should keep this for now. SPC has no business talking about non-PublicKeyCredentials at this point in time.

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
...
+                form the step above.
+
+            1. Return the tuple (|bbk|, |bbk_id|).
+
+            Note: Some platforms may provide identifiers for keypairs, while
+                others may provide a wrapped (encrypted) private key.
+                In either case the result can be presented as a buffer, and the
+                keypair handle subsequently retrieved from platform APIs.
+
+            Note: The user agent may provide a software implemention.
```
stephenmcgruer, line 1450: I think we need to open a (sub?) issue in GitHub tracking what we want to do for BBKs binding, and link it here. The options/parameters I know of for that issue are:

- Should we allow software at all? Or require hardware?
- Should we prefer hardware to software? (Your point below)
- Should we inform the RP about the storage type (I know in Chrome we don't want to do this, but we should have an issue tracking that discussion as its likely contentious)

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
...
+
+## Binding a KeyPair Internal Method ## {#sctn-binding-a-keypair}
+
+<dfn for="BrowserBoundKeyStore" method>BindKeyPair(bbk_id, credential_id)</dfn>:
+
+Given the |bbk_id|, the platform keypair identifier, and |credential_id|, the SPC credential_id (i.e. passkey id), stores the identifiers in the |browser_bound_map|.
+
+1. Insert the browser bound key id into the map. |browser_bound_map|[|credential_id|] = |bbk_id|.
+
+1. Serialize the map to disk.
```
stephenmcgruer, line 1467: Again avoid talking about how the map is stored.

(But talking about what to do in the algorithm if it fails is good. We can be generic, just like "if storing in the map fails, do X")

spec.bs
```diff
@@ -1165,6 +1369,173 @@ Note that there is no `paymentRequestOrigin` field in
...
+This component in the user agent manages platform-dependant cryptographic
+keypairs including their association to SPC credentials (i.e. passkeys). This
+section explains the |BrowserBoundKeyStore| and procedures for creating,
+binding, retrieving, and signing using the browser bound keypairs.
+
+The browser bound key store contains |browser_bound_map|, a map from
+[[=Credential ID=]] to a keypair identifier. The user agent stores
+this map on disk.
+
+## Creating a KeyPair Internal Method ## {#sctn-creating-a-keypair}
```
stephenmcgruer, line 1383: Probably more for your spec mentor to help with, but I'm not sure that these are internal methods as defined in https://tc39.es/ecma262/#sec-object-internal-methods-and-internal-slots

(And I'm not sure if an internal component is a concept at all?)

stephenmcgruer (commented):
Tried to keep this first pass to meta-level commentary, and sorry again for the delay in reviewing!

spec.bs
ianbjacobs: @stephenmcgruer, does the [requirements doc](https://github.com/w3c/secure-payment-confirmation/blob/main/bbk-requirements.md) answer those questions? I see this:

* To meet anticipated security requirements, issuance of the BBK by the user agent should involve a device-binding process that ensures a unique connection between the user agent (user profile) and the device. This may be, for example, through hardware crypto-security (e.g., TPM), keys stored in the secure element, or registration of the web browser linking a browser to a device.
* The user agent may return a BBK even in environments where a device-binding process is not readily available. 
* Each BBK should be associated with a signal indicating the nature of the device-binding process (e.g., corresponding to "secure element", "software", "no device binding").

Those have been socialized with the group.

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+
+<dl dfn-type="dict-member" dfn-for="CollectedClientAdditionalPaymentRegistrationData">
+    :  <dfn>browserBoundPublicKey</dfn> member
+    :: The base64url encoding of the [=browser bound key=] public key.
+</dl>
+
+# Browser Bound Key Store - <dfn interface>BrowserBoundKeyStore</dfn> Internal Component # {#sctn-browser-bound-key-store}
+
+This component in the user agent manages platform-dependant cryptographic
+key pairs including their association to SPC credentials (i.e. passkeys). This
```
stephenmcgruer, line 1408: Nit, link SPC credentials, e.g. `[=SPC credentials=]`.

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+section explains the {{BrowserBoundKeyStore}} and procedures for creating,
+binding, retrieving, and signing using the browser bound key pairs.
+
+Note: The BrowserBoundKeyStore needs to generate key pairs, retrieve key pairs,
+    export public keys, and generate signatures. Many operating system provide
+    APIs implement these operations and provide storage of the private in a
```
stephenmcgruer, line 1414: Typo: "APIs that implement"

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+    secure element on the device when available. For example [Android
+    KeyStore](https://source.android.com/docs/security/features/keystore),
+    [Apple CryptoKit](https://developer.apple.com/documentation/cryptokit/),
```
stephenmcgruer, line 1417: Grammar nit: ", or"

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+    [Windows Cryptography API: Next
+    Generation](https://learn.microsoft.com/en-us/windows/win32/seccng/cng-portal).
+
+The browser bound key store contains
+
+    : |browser_bound_map|
```
stephenmcgruer, line 1423: I think you want this to be a definition so that you can refer to it later, right? `<dfn>browser_bound_map</dfn>`, and then can use `[=browser_bound_map=]`.

(And for keypair_map)

spec.bs
stephenmcgruer: As long as they aren't attached to an interface, they have no impact on webpages. So it works just as is.

Often we refer to these as "steps to do XYZ", but thats not strictly required, I think it's just convention.

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+        Note: The user agent may be using a cryptographic API that provides an
+            equivalent of the |keypair_map| in which case the user agent would
+            delegate this aspect to the cryptographic API.
+
+## Creating a KeyPair Internal Method ## {#sctn-creating-a-keypair}
+
+<dfn for="BrowserBoundKeyStore" method>CreateKeyPair(allowed_algorithms)</dfn>:
+
+Given |allowed_algorithms| a {{PublicKeyCredentialParameters}} object from
+{{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}, creates a
```
stephenmcgruer, line 1438: I don't think you need to specify its from a {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}. Just something like:

`Given a {{PublicKeyCredentialParameters}}, |allowed_algorithms|, creates a ...`

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+key pair unless no allowed algorithms match, and returns a tuple of |bbk|, the
+key pair handle, and |bbk_id|, the key pair identifier along with a
+key pair identifier.
```
stephenmcgruer, line 1441: "the key pair identifier along with a key pair identifier."

I don't follow this. Maybe the "along with ..." was meant to be removed?

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+
+1. If the |allowed_algorithms| are an empty list, set this list to a
```
stephenmcgruer, line 1443: Nit: `|allowed_algorithm|` is a named variable (e.g., it can be treated as a proper noun), so you can just write:

`If |allowed_algorithms| is an empty list, set ...`

spec.bs
```diff
@@ -1159,12 +1377,202 @@ fields:
...
+        the relying party does not get the browser bound key on this
+        transaction. Or the user agent may continue in which case the browser
+        bound key is only used once. In either case, on a subsequent payment
+        assertion a new key pair will be created.
+
+## Retrieving a key pair Internal Method ## {#sctn-retrieving-a-browser-bound-key}
+
+<dfn for="BrowserBoundKeyStore" method>GetOrCreateBrowserBoundKey(credential_id, allowed_algorithms)</dfn>:
+
+Given |credential_id|, a [[=Credential ID=]], and |allowed_algorithms|, a
```
stephenmcgruer, line 1518: Just single square brackets for Credential ID, `[=Credential ID=]`

spec.bs
stephenmcgruer: Ack, thanks!

spec.bs
stephenmcgruer: Ack, sgtm. I think having this nested dictionary is good then.

spec.bs
pejic: I have created issue #288 to continue the discussion (and linked it here in the spec).

spec.bs
domfarolino: Exactly, you basically just want an algorithm whose title is a definition with some steps, and you can call that algorithm from any other place in the spec, including actual web API methods. See https://wicg.github.io/fenced-frame/#revoke-network-for-a-partition-nonce and https://github.com/WICG/fenced-frame/blob/master/spec.bs#L2216-L2226 for example. Let me know if that helps.

spec.bs
```diff
@@ -218,6 +228,20 @@ payment service providers are likely to build them on behalf of merchants.
...
+### Cryptographic evidence of device binding ### {#sctn-use-case-device-binding}
+
+In the payments industry a signal of device possession plays an important role
+as a second factor. WebAuthn allows for synced passkeys where one credential is
+available on multiple devices ([[webauthn-3#sctn-usecase-consumer-mdc]]).
+Although syncing improves the user experience for login use cases, concerns
+have been raised that synced passkeys alone do not satisfy device possession
+requirements in some regulatory environments.
+
+These concerns motivate the inclusion of an auxiliary [=key pair|public private
```
stephenmcgruer, line 240: Nit: either "of an auxiliary public private keypair" (singular) or "of auxiliary public private keypairs" (plural, no 'an').

spec.bs
```diff
@@ -468,6 +524,26 @@ try {
...
+    addition to the WebAuthn credential and is tied to a single device by the
+    user agent.
+
+: <dfn>Key Pair</dfn>
+:: A pair of asymmetric cryptographic keys. These are algorithm specific
+    parameters to the cryptographic algorithms referenced by the IANA COSE
+    Algorithms registry [[!IANA-COSE-ALGS-REG]].
+
+    The public key portion is returned during registration in
+    {{CollectedClientAdditionalPaymentRegistrationData}}/{{CollectedClientAdditionalPaymentRegistrationData/browserBoundPublicKey}}
```
stephenmcgruer, line 538: Nit; these should be period (`.`) separated, e.g. as in https://github.com/w3c/webauthn/blob/main/index.bs#L8490

```suggestion
    {{CollectedClientAdditionalPaymentRegistrationData}}.{{CollectedClientAdditionalPaymentRegistrationData/browserBoundPublicKey}}
```

(Here and elsewhere)

spec.bs
```diff
@@ -1045,7 +1139,63 @@ Note: Previously, the `payment` extension allowed for the creation of
               enrollment.https.html
             </wpt>
 
-:  <dfn export>Client extension processing ([=authentication extension|authentication=])</dfn>
+    1. Before step 13 (creating a {{CollectedClientData}}).
+
+        1. Let |bbk_allowed_algorithms| be {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}.
+
+            1. When
```
stephenmcgruer, line 1146:
```suggestion
            1. If
```

Optionally, can un-nest this step, so that one has:

```
1. Let bbk_allowed_algorithms be browserBoundPubKeyCredParams.
2. If bbk_allowed_algorithms is empty, let bbk_allowed_algorithms be PublicKeyCredentialCreationOptions.pubKeyCredParams.
3. Let (bbk - the platform specific key pair ...
```

(Either works, I like avoiding nesting, but others may prefer the visual 'scoping' of the nested 'if')

spec.bs
```diff
@@ -1045,7 +1139,63 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+                {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+                is empty, let |bbk_allowed_algorithms|
+                be
+                {{PublicKeyCredentialCreationOptions}}/{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+        1. Let (|bbk| - the platform specific key pair handle, |bbk_id| - an
+            identifier of the |bbk|) be the result of [=creating a key pair=]
+            using |bbk_allowed_algorithms|.
+
+            1. If [=create a key pair=] failed, skip steps below relating to
```
stephenmcgruer, line 1156: Super-nits:

- I think this should still be `creating a key pair` here, grammatically.
- "skip the steps below" (missing the definite article 'the' currently)

spec.bs
```diff
@@ -1045,7 +1139,63 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+                |bbk| and |bbk_public_key|.
+
+        1. [=Get a browser bound public key=] and let |bbk_public_key| be the result.
```
stephenmcgruer, line 1159: Nit: you need to pass in `bbk` here in some form, probably just needs to be `[=Get a browser bound public key=] for |bbk|`. Also probably flip the sentence to be:

```
1. Let |bbk_public_key| be the result of [=getting a browser bound public key=] for |bbk|.
```

spec.bs
```diff
@@ -1045,7 +1139,63 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+            :: |bbk_public_key| - The COSE_Key encoded public key.
+
+        1. All other fields set as per the original step 13.
+
+    1. In [=create credential async loop|step 22=], case "any authenticator
+        indicates success", step 3, just before returning <var
+        ignore>pubKeyCred</var>:
+
+        1. [=Bind a key pair=] using |bbk_id| and <var
+            ignore>pubKeyCred</var>.{{PublicKeyCredential/[[identifier]]}}.
```
stephenmcgruer, line 1177: @domfarolino - I'm not super used to internal slots, but looking at WebAuthn's spec, should this be worded as "and the <var ignore>pubKeyCred's</var> {{PublicKeyCredential/[[identifier]]}} internal slot." ?

(I.e., is it allowed to 'reference' internal slots via the `.` notation, or does one have to be more wordy as I've described above).

spec.bs
```diff
@@ -1045,7 +1139,63 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+            1. If [=binding a key pair=] failed, return an error,
+                {{UnknownError}}.
+
+        1. Let |payment_outputs| be {{AuthenticationExtensionsPaymentOutputs}} with:
+
+            : {{AuthenticationExtensionsPaymentOutputs/browserBoundSignature}}
+            :: A {{BrowserBoundSignature}} with fields:
+
+                : {{BrowserBoundSignature/signature}}
+                :: Result of [=generating a browser bound signature|generate a
```
stephenmcgruer, line 1188: I think this should be `generating`, not generate, i.e. drop the `|` and text after it.

spec.bs
```diff
@@ -1092,15 +1256,66 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+                    bound public key=].
 
-            1. All other fields set as per the original step 9.
+            1. All other fields set as per the original step 10.
 
             <wpt>
               authentication-accepted.https.html
             </wpt>
 
+        1. During "If any authenticator indicates success"; in step&nbsp; when
```
stephenmcgruer, line 1269: The step number is missing here

spec.bs
```diff
@@ -1092,15 +1256,66 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+            1. [=Bind a key pair=] using |bbk_id| and [=credentialIdResult=].
+
+            1. Let |payment_outputs| be
+                {{AuthenticationExtensionsPaymentOutputs}} with:
+
+                : {{AuthenticationExtensionsPaymentOutputs/browserBoundSignature}}
+                :: A {{BrowserBoundSignature}} with fields:
+
+                    : {{BrowserBoundSignature/signature}}
+                    :: Result of [=generating a browser bound
```
stephenmcgruer, line 1281: Again I think generating not generate here

stephenmcgruer (approved):
Some more minor nits, but overall this LGTM as a solid foundation for BBKs in the spec. There is still more to add, notably we need to add Security & Privacy Considerations for BBKs, but I think we can do that in a separate PR in the future.

Note that we should not land this yet. Firstly, I'm sure Dominic may have input as a spec mentor (thanks in advance Dominic!), as may Ian as a reviewer. However on top of that, we should give the Working Group the chance to review the content before this lands. @ianbjacobs - I would prefer not to wait until the 24th to do that, do you think it would suffice to send the preview to the Working Group mailing list and invite feedback that way?

(Even after this lands we can always come back and change it, of course!!)

Thank you Slobo for your work here!

spec.bs
```diff
@@ -1159,12 +1375,208 @@ fields:
...
+    [=credentialPublicKey=] from WebAuthn's
+    [[webauthn-3#sctn-attested-credential-data]].
+
+1. Return |encoded_public_key|.
+
+## Signing client data ## {#sctn-signing-client-data}
+
+To <dfn>generate a browser bound signature</dfn> given the |bbk| and
+|client_data_json|, a byte array containing the encoded client data perform the
+following steps:
```
ianbjacobs, line 1568: Does there need to be a comma after "data"? I wasn't sure how to parse this sentence otherwise.

ianbjacobs (approved):

spec.bs
pejic: Done.

spec.bs
pejic: Apparently I am still learning the bikeshed syntax: [=reference|displayed text=] instead of [=displayed text|reference=]. And it turns out the first word can be conjugated, so the displayed text portion is not needed.

spec.bs
pejic: Done.

spec.bs
pejic: Done here and at other instances of `}}/{{`.

spec.bs
pejic: Done (plural, no "an").

spec.bs
pejic: Done (when -> if).

Unnested here and 1 more location.

spec.bs
pejic: Yes there should be a comma after data to end the appositive to |client_data_json|. Thanks!

spec.bs
```diff
@@ -1022,6 +1109,13 @@ Note: Previously, the `payment` extension allowed for the creation of
       :  <dfn>instrument</dfn> member
       :: The instrument details that were displayed to the user. Only used at authentication time; not registration.
 
+      :  <dfn>browserBoundPubKeyCredParams</dfn> member
+      :: The list of allowed types of credential restricting the types of
+         cryptographic algorithms used for the browser bound key.
+
+         Note: When this member is not present, it defaults to
+             {{PublicKeyCredentialCreationOptions}}.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
```
domfarolino, line 1117: The indenting here seems weird? I don't think it affects the actual rendering, but it's still a little funky.

spec.bs
```diff
@@ -1045,7 +1139,62 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+        1. Let |bbk_allowed_algorithms| be {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}.
+
+        1. If
+            {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+            is empty, let |bbk_allowed_algorithms| be
+            {{PublicKeyCredentialCreationOptions}}.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+        1. Let (|bbk| - the platform specific key pair handle, |bbk_id| - an
+            identifier of the |bbk|) be the result of [=creating a key pair=]
+            using |bbk_allowed_algorithms|.
```
domfarolino, line 1153: This whole line doesn't really read correctly to me. If you have a non-normative note to attach to this line, put it in a Note. If you're trying to clarify the type, then just clarify the time without burying it in parens and dashes. I'm not sure what "platform specific key pair handle" means, but it seems like you might need https://infra.spec.whatwg.org/#implementation-defined.

spec.bs
```diff
@@ -1159,12 +1373,208 @@ fields:
...
+
+To <dfn>bind a key pair</dfn> given the |bbk_id|, the key pair identifier, and
+|credential_id|, the SPC credential_id (i.e. passkey id), store the
+identifiers in the [=browser_bound_map=] as follows:
+
+1. Insert the browser bound key id into the map:
+    [=browser_bound_map=][|credential_id|] = |bbk_id|.
+
+    Note: When user agents are storing this map (such as serializing to disk)
+        and the procedure fails, the user agent may return an error
```
domfarolino, line 1512: This is normative, so it cannot be buried inside a note. You need an explicit step saying to handle any implemetation-defined errors by returning "failure" or returning null, or something else that the callers can explicitly key their behavior off of. But the point is, you have to account for that failure case explicitly in this algorithm.

spec.bs
```diff
@@ -1159,12 +1373,208 @@ fields:
...
+        :: Stop, returning null.
+
+            Note: The user agent does not add any browser bound outputs in this
+            case.
+
+        : is non-zero
+        :: Proceed to create the key pair:
+
+            1. Let |chosen_algorithm| be the first entry of
+                |allowed_algorithms|.
```
domfarolino, line 1470: `Let |chosen_algorithm| be |allowed_algorithms][0].`

spec.bs
```diff
@@ -1159,12 +1373,208 @@ fields:
...
+
+            1. Generate the public private key pair using the established
+                procedure corresponding to |chosen_algorithm|.
+
+            1. If the step above failed for any reason, return null.
+
+            1. Let |bbk| be the the key pair from the key generation step above.
+
+            1. Generate an identifier, |bbk_id|.
```
domfarolino, line 1479: Same here.

spec.bs
```diff
@@ -1159,12 +1373,208 @@ fields:
...
+            1. Bind the key pair to the SPC credential by [=binding a key
+                pair=] using |bbk_id| and |credential_id|.
+
+    </dl>
+
+1. Return |bbk|.
+
+## Getting the browser bound public key ## {#sctn-getting-the-bbk-public-key}
+
+To <dfn>get a browser bound public key</dfn> given the |bbk|, the key pair
```
domfarolino, line 1550: Specify the type of all inputs on all algorithms.

spec.bs
```diff
@@ -1159,12 +1373,208 @@ fields:
...
+
+## Signing client data ## {#sctn-signing-client-data}
+
+To <dfn>generate a browser bound signature</dfn> given the |bbk| and
+|client_data_json|, a byte array containing the encoded client data, perform
+the following steps:
+
+1. Generate |signature|, a byte array, by performing the cryptographic
+    signature algorithm on |client_data_json|. This corresponding algorithm was
+    chosen when [=creating a key pair=].
```
domfarolino, line 1570: You have to clarify where the "cryptographic signature algorithm" comes from. Either pass it in, store it somewhere outside this algorithm that this algorithm has access to, etc. It's too "magical" as-is.

spec.bs
pejic: Done.

spec.bs
pejic: I have foregone clarifying the types here: The types are stated in the Browser Bound Key Store section (including implementation defined as needed).

pejic (commented):
Thanks for the feedback @domfarolino. Could you take another look?

spec.bs
```diff
@@ -1159,12 +1375,248 @@ fields:
...
+    #288](https://github.com/w3c/secure-payment-confirmation/issues/288) that
+    discusses storage type of BBKs including additional topics of which storage
+    types should be allowed and exposing the storage type to the relying party.
+
+## Binding a key pair ## {#sctn-binding-a-keypair}
+
+To <dfn>bind a key pair</dfn> given the |bbk_id|, the byte array containing the
+key pair identifier, and |credential_id|, the byte array of the SPC
+credential_id (i.e. passkey id), store the identifiers in the
+[=browser_bound_map=] as follows:
```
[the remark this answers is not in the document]
domfarolino: I think I would put the Note in between the algorithm declaration, and the body/steps.

spec.bs
```diff
@@ -1045,7 +1144,61 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+        1. All other fields set as per the original step 13.
+
+    1. In [=create credential async loop|step 22=], case "any authenticator
+        indicates success", step 3, just before returning <var
+        ignore>pubKeyCred</var>:
+
+        1. [=Bind a key pair=] using |bbk_id| and <var
+            ignore>pubKeyCred</var>.{{PublicKeyCredential/[[identifier]]}}.
+
+        1. If [=binding a key pair=] failed, return an error, {{UnknownError}}.
```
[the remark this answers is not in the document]
pejic: Thanks. Leaving this as-is, i.e., don't return non-failure.

spec.bs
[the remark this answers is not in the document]
pejic: Thanks, rearranged so the Note comes between the declaration and steps.

spec.bs
```diff
@@ -483,6 +546,30 @@ try {
     [[webauthn-conditional-ui inline]]. It is likely that both it and SPC could
     be supported by the same underlying API.
 
+: <dfn>Browser bound key</dfn>
+:: A public-private [=key pair=] that signs over the transaction details in
+    addition to the WebAuthn credential and is tied to a single device by the
+    user agent.
+
+: <dfn>Key Pair</dfn>
+:: A key pair is a [=struct=] consisting of <dfn>public key</dfn> and
```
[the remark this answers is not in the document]
pejic: Done. The refs still work.

spec.bs
```diff
@@ -483,6 +546,30 @@ try {
...
+    <dfn>private key</dfn>, implementation defined types.
```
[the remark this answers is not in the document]
pejic: Done. Here and another instance.

spec.bs
```diff
@@ -1257,7 +1362,60 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+        1. If the result of [=creating a key pair=] is null, skip the steps
+            below relating to |bbk_and_algorithm| and |bbk_public_key|.
+
+        1. Let |bbk_public_key| be the result of [=getting a browser bound
+            public key=] using |bbk_and_algorithm|.
+
+    1. In step 13, instead of creating the {{CollectedClientData}} create a
+        {{CollectedClientPaymentData}} whose fields are:
+
+        1. {{CollectedClientPaymentData/payment}} set to a
```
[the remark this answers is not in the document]
domfarolino: Sounds good to me.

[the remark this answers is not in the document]
pejic: Done.

spec.bs
```diff
@@ -1257,7 +1362,60 @@ Note: Previously, the `payment` extension allowed for the creation of
...
+
+        1. If
+            {{AuthenticationExtensionsPaymentInputs/browserBoundPubKeyCredParams}}
+            is empty, let |bbk_allowed_algorithms| be
+            {{PublicKeyCredentialCreationOptions}}.{{PublicKeyCredentialCreationOptions/pubKeyCredParams}}.
+
+        1. Let |bbk_and_algorithm|, |bbk_id| be the result of [=creating a key
+            pair=] using |bbk_allowed_algorithms|.
+
+        1. If the result of [=creating a key pair=] is null, skip the steps
```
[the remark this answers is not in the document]
pejic: Done.

spec.bs
```diff
@@ -93,6 +102,8 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
 spec:fetch; type:dfn; for:/; text:request;
 spec:i18n-glossary; type:dfn; text:bidi isolation
 spec:i18n-glossary; type:dfn; text:language priority list
+spec:infra; type:dfn; text:list
+spec:infra; type:dfn; text:user agent
```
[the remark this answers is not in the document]
pejic: The `list` and `user agent` are already in the link defaults block (starts on line 95 in the latest version), so I think this is good as it stands. However, I could be misunderstanding.

domfarolino (approved):

pejic:
Thank you Dominic, Ian, and Stephen for the reviews on Browser Bound Keys!

PR #293, opened 2025-05-06
Adjustments to BBK requirements

ianbjacobs:
Based on recent conversations:

* Clarify that a BBK should not be usable outside of the device where it was originally bound.
* Make explicit the implication that if a passkey is deleted then any associated BBK should also be deleted.


Goosth:
Do we need to link BBK's to a specific passkey? I feel it raises a number of other questions:
* What does this mean if a Passkey is used on a different device and removed there?
* What about a new passkey being issued, since the browser/RP could not detect the previous Passkey anymore (cookie was deleted).
It would be great to leave open the option of having a BBK in the future which we only want to use that as a possession factor without the direct link to Passkeys. So basically the BBK would work more along the principles of Server Side Cookies with CHIPS.

ianbjacobs:
I have updated this pull request following the [8 May 2025 WPWG meeting](https://www.w3.org/2025/05/08-wpwg-minutes.html#3024).

bbk-requirements.md
```diff
@@ -16,50 +16,61 @@ Therefore, the Web Payments Working Group plans to add a "browser-based key (BBK
...
 
-* If a BBK is deleted, the user agent should generate a new BBK associated with the same passkey.
+* Each BBK should be associated with a signal indicating the nature of the device-binding process (e.g., corresponding to "secure element", "software", "no device binding"). <b>Status Note:</b> As of May 2025, this feature is not yet defined in the SPC
+specification, nor has it been implemented in Chrome.
 
-* At authentication time, the client data will be signed by the passkey and the associated BBK.
+### Association with passkeys
 
-* To link the BBK to the passkey cryptographically, the BBK public key should be added in the client data.
+* At both passkey generation time and SPC authentication time, if there is no BBK associated with a [BBK binding](#bbk-binding), the user agent should create one and associate it with that [BBK binding](#bbk-binding).
```
pejic, line 49: At passkey generation time, should we require an indication that this passkey is used for payments? If not every passkey (even those never intended for use with payments) would need to have a BBK created for them.

Currently we have the [`isPayment`](https://www.w3.org/TR/secure-payment-confirmation/#dom-authenticationextensionspaymentinputs-ispayment) member on the payment extensions input (which functions effectively as a third-party payment enablement since passkeys created with `isPayment = false` can still be used in the first-party context).

bbk-requirements.md
```diff
@@ -16,50 +16,61 @@ Therefore, the Web Payments Working Group plans to add a "browser-based key (BBK
...
+* The consensus of the Web Payments Working Group is that, initially, each BBK will be paired with a passkey. We assume here that a BBK will be bound to a single passkey. In the future the Web Payments Working Group may explore the use of BBKs without passkeys.
+* A given passkey corresponds to an online identity. To distinguish online identities, the user will use different passkeys. 
 * A relying party will perform some ID &amp; V process before trusting a (new) BBK. That ID &amp; V process might take place before a Web Authentication registration (and thus, if the BBK is returned as part of the Web Authentication registration, the RP would not likely step up the user a second time). In the case of a synched passkey, when the RP first sees a BBK on a new user device, in the absence of other trust signals, the RP would likely perform some ID &amp; V process in order to trust the new BBK, and we consider that an acceptable user experience on a new device.
 
+## Definitions
+
+* <dfn id="bbk-binding">BBK binding</dfn>: a unique triplet consisting of a specific passkey, a specific browser instance, and a specific device.
```
pejic, line 27: Regarding "a specific browser instance", does this leave room for interpretation relating to profiles? Profiles being roughly browser's way of distinguishing a different user, so that both of these scenarios would meet requirements:

- browsers that create a BBK per browser process and per profile (and still per passkey, per device), and
- browsers that create a BBK per browser process and *not* per profile (and still per passkey, per device).

(Allowing for both was the motivation for amending the specification as I recall: [May 8th minutes](https://www.w3.org/2025/05/08-wpwg-minutes.html))

bbk-requirements.md
```diff
@@ -16,50 +16,61 @@ Therefore, the Web Payments Working Group plans to add a "browser-based key (BBK
...
+* At both passkey generation time and SPC authentication time, if there is no BBK associated with a [BBK binding](#bbk-binding), the user agent should create one and associate it with that [BBK binding](#bbk-binding).
 
-### User agent user profiles
+* Once the user agent has associated a BBK with a [BBK binding](#bbk-binding), the user agent should use that BBK whenever the relevant passkey is used with SPC authentication on this device. 
 
-* A user agent may reuse the same BBK across user profiles of the same user agent instance.
+* At SPC authentication time, the client data will be signed by the passkey and the associated BBK.
 
-### Device binding
+* To link the BBK to the passkey cryptographically, the BBK public key should be added in the client data.
```
pejic, line 55: The passkey must sign over the BBK public key otherwise the third-party page generating the BBK (instead of the browser) is not distinguishable to the relying-party.

```suggestion
* To link the BBK to the passkey cryptographically, when the user agent does provide a BBK signature, the BBK public key must be added in the client data.
```

bbk-requirements.md
```diff
@@ -16,50 +16,61 @@ Therefore, the Web Payments Working Group plans to add a "browser-based key (BBK
...
+### Association with passkeys
 
-* To link the BBK to the passkey cryptographically, the BBK public key should be added in the client data.
+* At both passkey generation time and SPC authentication time, if there is no BBK associated with a [BBK binding](#bbk-binding), the user agent should create one and associate it with that [BBK binding](#bbk-binding).
 
-### User agent user profiles
+* Once the user agent has associated a BBK with a [BBK binding](#bbk-binding), the user agent should use that BBK whenever the relevant passkey is used with SPC authentication on this device. 
 
-* A user agent may reuse the same BBK across user profiles of the same user agent instance.
+* At SPC authentication time, the client data will be signed by the passkey and the associated BBK.
```
pejic, line 53: Should we also include at registration time when the passkey is intended for payment use cases? (In the spec this would be when `isPayment = true` in the payment extension inputs). This alleviates needing to re-authenticate on the first payment. Example scenario:

1. user registers a passkey after authenticating with an issuer,
2. a bit later user agrees to pay using SPC,
3. issuer sees the BBK for the first time and thus decides to step up the transaction.


I have been operating on the assumption that this was a requirement, but wasn't able to find a concrete yes in the minutes: There was discussion in [November 07 minutes](https://www.w3.org/2024/11/07-wpwg-minutes) "… from our perspective we would strongly prefer BBK creation at WebAuthn registration time after an ID&V." . And recently mentioned mention of this in the minutes [March 27 minutes](https://www.w3.org/2025/03/27-wpwg-minutes.html) "… they will get a BBK on creation, and that's the one they want to trust".

bbk-requirements.md
ianbjacobs: Adopted.

bbk-requirements.md
ianbjacobs: Yes, please let's also include at registration time (but without the need to know whether the passkey will be used for payments). Can we simply edit the sentence to say "At both SPC registration and authentication time, ..."?

ianbjacobs: Made the change.

bbk-requirements.md
ianbjacobs: In discussion at the [22 May 2025 WPWG meeting](https://www.w3.org/2025/05/22-wpwg-minutes.html) I did not hear a push from the Working Group to more rigorously define "browser instance." I think we can resolve this conversation.

bbk-requirements.md
ianbjacobs: In [discussion at the 22 May 2025 WPWG meeting](https://www.w3.org/2025/05/22-wpwg-minutes.html#7403) it was observed that:

* If a passkey does not have the payment bit set, at **registration time** there's no obvious way to return a BBK. I think that we can resolve this bit of conversation (but update the requirements accordingly) that unless a passkey has a payment bit set, there will be no BBK generated at registration time.
* We generalized the question to "Should SPC only be usable with passkeys that have the payment bit set?" See the new #299. One question asked: if an RP wants a passkey originally created for login to be usable with SPC, is there a way to "upgrade" the credential? The Chrome Team said they would look into this.

ianbjacobs:
@pejic, I have updated the pull request to say that at registration time, if the passkey has a payment extension, then the browser should create a BBK if there isn't one already. I did not go so far as to say "If there is no payment extension, the brower must not generate a BBK." I didn't think that was necessary, but if you think we should have strong language there, let me know.

Issue #197, opened 2022-08-08
Add issuer and payment system logos to transaction dialog


ianbjacobs:
Hi all,

Our EMVCo colleagues working on 3DS have said they would like to see the transaction dialog further aligned with the 3DS UI template. I have heard the following comments based on EMVCo UX studies:

* It is important for the cardholder to understand who is performing the authentication
* It is important to have a repetitive user experience and interface during the challenge

Because of the difference between the SPC transaction dialog and 3DS UI template, I have heard concerns that the cardholder may not trust this UI, which could lead to abandonments, or worse to a loss of trust in 3DS.

The concrete suggestion I have heard is to add the issuer and payment system logos to the SPC transaction dialog.

apouillard:
Hi all,

I can confirm that we have heard from issuing banks we are working with that they would like to be clearly identifiable by the cardholder when using the SPC UI just like it is the case today with the 3DS UI template.

mountainhippo:
I think this is a good idea as long as we keep it generalised i.e we have a
relying party logo.

It would be a very sad thing to constrain SPC to only work with 3DS/Cards



--
Nick Telford-Reed
e: ***@***.***
m: +447538177619

adrianhopebailie:
> the 3DS UI template

Can we get a link to this?

ianbjacobs:
@adrianhopebailie, see this presentation from Sep 2023:
 http://www.w3.org/2022/Talks/emvco-tpac-2022.pptx

Goosth:
EMVCo published a set of UX guidelines:

  *   https://3ds-ux-guidelines.emvco.com/

Not sure if this has yet been updated for 3DS 2.3, but it's a good start.

ianbjacobs:
From a recent conversation, just to be clear: these extra logos would be optional to provide.

ianbjacobs:
Thanks @darwinlyang!

I would like to hear from others about the value and scalability of the PaymentIconType enum. I foresee a couple of issues:

* I think (but am not 100%) that the terms "issuer" and "network" are card-centric, and it might create confusion for developers using SPC in other contexts.
* I imagine that the ultimate goal of the developer is to "get the icons to show up where I want them." To make it easier to achieve that goal, I think the "ordering of icons" in the input data more directly speaks to the end goal, provided that the specification also sets some expectation about the impact of order on display. I'm not sure that the abstraction of issuer/network clarifies enough to the developer how to achieve their end goal.
 
[...]

PR #294, opened 2025-05-08
[Spec] Add support for displaying logos from facilitating payment entities

stephenmcgruer:
This change adds the ability for the SPC caller to specify a sequence of 'facilitating entity logos', which are intended to be logos from payment entities (e.g., card network or issuer in the credit card space) that can provide context for the authentication to the user. Logos are specified by the caller in a descending order of importance.

In order to avoid constraining individual user agent interface design, the user agent is free to show a subset (including the empty subset) of the input logos, but it must respect the order of importance - that is, it can only show a logo with index j in the list if it is also showing all logos with indices i < j.

This change is intended to be future compatible for a world where we may later add additional information about the logos, including potential rendering hints and accessibility information, but for now constrains itself to the minimal possible support.

See #197

Tasks:

- [x] Chromium bug for implementation - https://crbug.com/417683819
- [ ] WPT tests added/updated - https://github.com/web-platform-tests/wpt/pull/53358

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/294.html" title="Last updated on May 23, 2025, 5:22 PM UTC (eaaa777)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/294/72eb74e...eaaa777.html" title="Last updated on May 23, 2025, 5:22 PM UTC (eaaa777)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1037,6 +1037,15 @@ The [=steps to respond to a payment request=] for this payment method, for a giv
     : {{AuthenticationExtensionsPaymentInputs/paymentEntitiesLogos}}
     :: |data|["{{SecurePaymentConfirmationRequest/paymentEntitiesLogos}}"] if
         it is present and non-empty, otherwise omitted.
+
+        Issue: The current version of this specification supports only a single
+        URL per {{PaymentEntityLogo}}, so simply copying and signing the data
+        structure suffices to indicate what was shown to the user. A future
+        version of this specification may add multiple URLs per
+        {{PaymentEntityLogo}} in order to, e.g., natively support dark mode. If
+        this happens, the specification will need to indicate to the [=Relying
+        Party=] which URL was shown for a given {{PaymentEntityLogo}}.
+
     : {{AuthenticationExtensionsPaymentInputs/total}}
     :: |request|.[=payment request details|[[details]]=]["{{PaymentDetailsInit/total}}"]
     : {{AuthenticationExtensionsPaymentInputs/instrument}}


spec.bs
```diff
@@ -579,6 +584,7 @@ NOTE: This allows the user agent to not require user activation, for
         unsigned long timeout;
         USVString payeeName;
         USVString payeeOrigin;
+        sequence<FacilitatingEntityLogo> facilitatingEntityLogos;
```
stephenmcgruer, line 587: Would love to figure out the best name here.

- `facilitatingEntityLogos` ?
- `facilitatingEntitiesLogos` ?
- Or something else entirely?!

stephenmcgruer:
cc @darwinlyang

spec.bs
ianbjacobs: I had wondered about paymentSystemLogos, but that appears to have a specific meaning in 3DS ("Payment System Image") and so may not be as generic-sounding as I had hoped.

I wonder if paymentEntityLogos is slightly easier to digest.

spec.bs
```diff
@@ -846,6 +870,25 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
     matches, to defeat attempts to [[#sctn-privacy-probing-credential-ids|probe
     for credential existence]].
 
+1. Optionally, the user agent may remove entries from
+    |data|["{{SecurePaymentConfirmationRequest/facilitatingEntityLogos}}"],
+    starting from the back of the list and moving backwards.
+
+    Note: This allows the user agent to control how many logos it will display,
+    while preserving the semantics that logos are in decreasing order of
+    importance to the caller.
```
ianbjacobs, line 879: Align "importance" here with whatever ends up being said above. For example: "in decreasing order of display priority to the caller."

spec.bs
```diff
@@ -1350,6 +1438,13 @@ Confirmation, the [=Relying Party=] MUST proceed as follows:
         * Verify that the value of |C|["{{CollectedClientPaymentData/payment}}"]["{{CollectedClientAdditionalPaymentData/payeeOrigin}}"]
             matches the origin of the payee that should have been displayed to the user, if any.
 
+        * Verify that the value of |C|["{{CollectedClientPaymentData/payment}}"]["{{CollectedClientAdditionalPaymentData/facilitatingEntityLogos}}"]
+            is a strict and ordered subset of the logos that should have been displayed to the user, if any.
+
+            Note: The user agent is permitted to not display all logos, but must
+            not include logos in the {{CollectedClientAdditionalPaymentData}}
+            that were not shown to the user.
```
ianbjacobs, line 1446:
```suggestion
           The user agent is not required to display all logos. The user agent must
            not include logos in the {{CollectedClientAdditionalPaymentData}}
            that were not shown to the user.
```

ianbjacobs: Because there is a normative requirement here ("must  not") this should not be a Note.

spec.bs
```diff
@@ -619,6 +625,12 @@ members:
         merchant). Optional, may be provided alongside or instead of
         {{SecurePaymentConfirmationRequest/payeeName}}.
 
+    :  <dfn>facilitatingEntityLogos</dfn> member
+    :: An optional list of logos representing entities that are facilitating
+        the payment that this SPC call is for, in descending order of
+        importance. The user agent is not required to show any or all of the
+        logos, see the [[#sctn-steps-to-check-if-a-payment-can-be-made]].
+
```
ianbjacobs, line 633: I think it could be useful here to set some expectations for the caller that, although order matters, the specification is intentionally not requiring a particular layout of the logos. For example:

Note: It is intentional that this specification does not specify exactly how logos will be rendered so that the user agent can make a best effort in light of many considerations including the payment method, device considerations, and logo dimension considerations. The order of logos in the list does matter, however, which means that developers can experiment to determine whether a particular order gives desired results.

spec.bs
```diff
@@ -619,6 +625,12 @@ members:
         merchant). Optional, may be provided alongside or instead of
         {{SecurePaymentConfirmationRequest/payeeName}}.
 
+    :  <dfn>facilitatingEntityLogos</dfn> member
+    :: An optional list of logos representing entities that are facilitating
+        the payment that this SPC call is for, in descending order of
+        importance. The user agent is not required to show any or all of the
```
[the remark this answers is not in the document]

spec.bs
stephenmcgruer: Done, thanks! Slightly modified from your suggested text so please check it over.

spec.bs

spec.bs
stephenmcgruer: Good catch! Even when not a note this might still not be ok, because this section is not really part of the user agent's specification - it's the "SPC Relying Party Operations".

Possibly we should drop this entirely, or keep it as a note but reword to not have `must` and make it informative only?

ianbjacobs: Does CollectedClientAdditionalPaymentData include these facilitated entity logos? It looks like it only includes *instrument* information. (And the WebIDL suggest this is not a list.)

So how do facilitated entity logos make it into the signed data? If that happens somewhere, then whereever that is specified, it should be a requirement to only include displayed logos.

Then the SPC RP Operations can include a note that the signed data will only include the logos actually displayed.

spec.bs
```diff
@@ -846,6 +877,25 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
...
+
+    Note: This allows the user agent to control how many logos it will display,
+    while preserving the semantics that logos are in decreasing order of
+    display priority to the caller.
+
+1. For each |logo| in |data|["{{SecurePaymentConfirmationRequest/facilitatingEntityLogos}}"]:
+
+    1. [=fetch an image resource|Fetch the image resource=] for the |logo|, passing
+        «["{{ImageResource/src}}" → |logo|["{{FacilitatingEntityLogo/url}}"]»
+        for *image*. If this fails, remove |logo| from
```
pejic, line 892: Above developer given some control via the order: "Note: ... The order of logos in the list does matter, however, which means that developers can experiment to determine whether a particular order gives desired results."

Removing logo here changes the index of subsequent successfully-downloading logos. If the web developer intentionally put logoX in position 3 (which, say, appears in the bottom right), but logoY in position2 (which, say, appears in the top left) fails to download, then logoX would end up being displayed in the top left rather than bottom right.

The user agent is free to display the icons in any manner, and the common case might be a row of icons where the missing logos would simply be skipped as if fewer logos were provided; however, if we are trying to give developers some control via the order then the list here should have some placeholder value, so that user agents with more complex logo placement are free to show missing logos at the UI level in another manner. (One could imagine a user agent showing a missing image icon instead of removing the image from a row of images.)

spec.bs
```diff
@@ -846,6 +877,25 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
...
+    |data|["{{SecurePaymentConfirmationRequest/facilitatingEntityLogos}}"],
+    starting from the back of the list and moving backwards.
+
+    Note: This allows the user agent to control how many logos it will display,
+    while preserving the semantics that logos are in decreasing order of
+    display priority to the caller.
+
+1. For each |logo| in |data|["{{SecurePaymentConfirmationRequest/facilitatingEntityLogos}}"]:
+
+    1. [=fetch an image resource|Fetch the image resource=] for the |logo|, passing
```
pejic, line 890: Is failure to decode the image a failure of the image resource fetcher, or will we need to mention that here?

spec.bs
ianbjacobs: I think it would be useful to add a note that the display order may be affected when logos fail to download successfully.

spec.bs
stephenmcgruer: > Does CollectedClientAdditionalPaymentData include these facilitated entity logos? It looks like it only includes instrument information. (And the WebIDL suggest this is not a list.)

It does include them - that is added in this change too :). See https://github.com/w3c/secure-payment-confirmation/pull/294/files#diff-6f5a1d8263b0b0c42e2716ba5750e3652e359532647ac934c1c70086ae3ceddaR1217 (or step 2.2.2 of https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/294.html#client-extension-processing-authentication)

spec.bs
darwinlyang: +1 I like `paymentEntityLogos` or `paymentEntityIcons` more.

spec.bs
stephenmcgruer: Good question! I spent a while digging and I don't think that decoding is part of the image resource fetcher algorithm, nor can I find much of a useful definition outside of the HTML spec - and it's a bit HTML-specific in that case (https://html.spec.whatwg.org/multipage/images.html#img-decoding-process).

For now, I have updated this to say "Fetch the image resource for the logo, ..., and decode it. If either fetching or decoding fails ...", and chosen to not define what decoding means.

spec.bs
stephenmcgruer: Updated to `paymentEntitiesLogos`, since it contains the logos for more than one payment entity. Open to bikeshedding whether it should still be singular just for readability/ease of spelling.

cc @domfarolino to ask - Dom, do you know if there is any spec authors guidance on naming cases like this (a sequence of logos from multiple different payment entities, is that called `paymentEntityLogos` or `paymentEntitiesLogos`...?)

spec.bs
stephenmcgruer: Good catch! I've followed what we do for the instrument icon and just cleared the URL instead of removing the entry.

spec.bs
```diff
@@ -1083,6 +1145,9 @@ Note: Previously, the `payment` extension allowed for the creation of
       :  <dfn>payeeOrigin</dfn> member
       :: The payee origin, if present, that was displayed to the user. Only used at authentication time; not registration.
 
+      :  <dfn>paymentEntitiesLogos</dfn> member
+      :: The logos, if any, that were displayed to the user representing entities that are facilitating the payment that this SPC call is for. Only used at authentication time; not registration.
```
pejic, line 1149: The user agent can scale the images to fit. If someone creates a 1024x2048 image (say filled with additional terms) should they expect that having been signed over her in CollectedClientAdditionalPaymentData that the user has seen all the detail of the image? Should we include dimensions or scale so it can be know after the fact what the user could have seen, or state somewhere that the image are shown on a best effort basis - scaling to accommodate the UI?

(We don't do this for the payment instrument icon, so maybe it was discussed and decided on previously).

rsolomakhin:
Hello,

May I suggest that the API require a non-empty accessibility label for SPC icons? This seems like a fairly important property for accessibility tools. 

Although we can't control that websites put something sensible in this label, the other options (either no accessibility label or a generic label from the user agent, like "icon 1") may reduce accessibility of this UI with screen readers.

Cheers,\
Rouslan

stephenmcgruer:
Thanks for raising that Rouslan - it's a good point that if we want a field to be `required`, it has to be added 'now' or it won't be backwards/forwards compatible to make that change in the future. So if we think a `label` field is required for accessibility, then we should include it in this CL.

@ianbjacobs - just to check, are you aware of any recommendations accessibility horizontal review folks on the right way to spec this? Just would be nice to be consistent with whatever is out there (e.g., is the name `label` reasonable, is there any guidance we should give callers, should we let the user agent truncate the label if it wants or even write that into the spec, etc).

ianbjacobs:
> @ianbjacobs - just to check, are you aware of any recommendations accessibility horizontal review folks on the right way to spec this? Just would be nice to be consistent with whatever is out there (e.g., is the name `label` reasonable, is there any guidance we should give callers, should we let the user agent truncate the label if it wants or even write that into the spec, etc).

We can ask for review of this pull request; I will check with my colleagues.

Related: given the large number of recent spec enhancements, we will want to discuss when we seek updated horizontal review of the specification.

iadawn:
> May I suggest that the API require a non-empty accessibility label for SPC icons? This seems like a fairly important property for accessibility tools.

+1 to this suggestion. Might also be worth considering adding the same for `PaymentEntityLogo`.

If you want APA to take a look at this before any formal review request, give me a shout and I can see if they have bandwidth.

rsolomakhin:
> Might also be worth considering adding the same for `PaymentEntityLogo`.

I had `PaymentEntityLogo` in mind for [my accessibility logo comment.](https://github.com/w3c/secure-payment-confirmation/pull/294#issuecomment-2876817191)

@iadawn - Is there another logo that you also were thinking of, e.g., in [PaymentCredentialInstrument](https://w3c.github.io/secure-payment-confirmation/#dictdef-paymentcredentialinstrument)?

Although the [PaymentCredentialInstrument](https://w3c.github.io/secure-payment-confirmation/#dictdef-paymentcredentialinstrument) dictionary contains an `icon` field, it also has a `displayName` field, which can be used for an accessibility announcement. Because of that, I don't feel like there is a need for an additional accessibility label for that `icon`, but I would like to hear your opinion on this as well.

iadawn:
Ah no, I think I misread something and thought there was a second icon. All good :)

stephenmcgruer:
Ok thanks folks, I've added the `label` into the spec. See the latest commit in this PR (https://github.com/w3c/secure-payment-confirmation/pull/294/commits/cdf3185161086a135e755cbde9db1c107ed545dd) or the preview (https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/294.html#dom-paymententitylogo-label). Feedback very welcome :)

spec.bs
stephenmcgruer: (Resolving this, please re-open if you are still concerned!)

spec.bs
stephenmcgruer: I had a chat with our standards folks on this, and after some debate nobody could come up with something that was clearly better than this (everything has trade-offs), and we felt that this was 'good enough' and won't cause (too many) problems as-is (`paymentEntitiesLogos`). So resolving this conversation.

spec.bs
```diff
@@ -1062,6 +1123,7 @@ Note: Previously, the `payment` extension allowed for the creation of
       USVString topOrigin;
       USVString payeeName;
       USVString payeeOrigin;
+      sequence<PaymentEntityLogo> paymentEntitiesLogos;
```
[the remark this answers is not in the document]
stephenmcgruer: In terms of deferring this decision, the current setup technically does that, right? As long as we don't add new fields to `PaymentEntityLogo`, it's forwards-compatible with a future split off to some `PaymentEntityLogoShown` dictionary - as long as we assume the minimal elements are the `url` and `label`.

That said, it might still be dangerous if we then accidentally add (e.g.) dark mode urls to `PaymentEntityLogo` and fail to make the decision whether or not to split the output at that time, and so accidentally just end up with output==input without indicating which was chosen.

Perhaps just filing an issue to track that, and linking to it in the spec will suffice? Otherwise I'm happy to introduce `PaymentEntityLogoShown` now.

spec.bs
stephenmcgruer: This wasn't explicitly discussed, but we did previously discuss how deep we wanted to go. For example, if:

1. I provide the icon URL `https://stephen.example/icon.png`
2. You get a cryptogram today that specifies that URL
3. Tomorrow, I change the bytes that the server serves to be different

At the time we talked about hashing the image and signing that hash, etc etc, but ultimately decided it was just too much effort for something that isn't *that* key (vs the actual transaction details).

So I would be tempted to lean on that prior art and just say "this isn't important enough to try to address". What do you think?

spec.bs
```diff
@@ -846,6 +886,31 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
     matches, to defeat attempts to [[#sctn-privacy-probing-credential-ids|probe
     for credential existence]].
 
+1. Optionally, the user agent may remove entries from
+    |data|["{{SecurePaymentConfirmationRequest/paymentEntitiesLogos}}"],
+    starting from the back of the list and moving backwards.
```
ianbjacobs, line 891:
```suggestion
    starting from the end of the list and moving towards the start of the list.
```

spec.bs
```diff
@@ -903,6 +968,14 @@ authentication:
     {{PaymentCredentialInstrument/iconMustBeShown}} must be `false` here as
     otherwise the [[#sctn-steps-to-check-if-a-payment-can-be-made|the steps
     to check if a payment can be made]] would have failed previously.
+* The logos represented by the {{PaymentEntityLogo/url}} entries from
+    {{CollectedClientAdditionalPaymentData/paymentEntitiesLogos}}, if
+    non-empty.
+
+      * The user is not required to display the {{PaymentEntityLogo/label}}
```
ianbjacobs, line 975:
```suggestion
      * The user agent is not required to display the {{PaymentEntityLogo/label}}
```

spec.bs
```diff
@@ -1083,6 +1160,9 @@ Note: Previously, the `payment` extension allowed for the creation of
       :  <dfn>payeeOrigin</dfn> member
       :: The payee origin, if present, that was displayed to the user. Only used at authentication time; not registration.
 
+      :  <dfn>paymentEntitiesLogos</dfn> member
+      :: The logos, if any, that were displayed to the user representing entities that are facilitating the payment that this SPC call is for. Only used at authentication time; not registration.
```
ianbjacobs, line 1164:
```suggestion
      :: The logos, if any, that were displayed to the user in the transaction dialog. These logos are intended to represent the entities facilitating this transaction. Only used at authentication time; not registration.
```

spec.bs
```diff
@@ -1221,6 +1305,9 @@ fields:
     :  <dfn>payeeOrigin</dfn> member
     :: The origin of the payee, if present, that was displayed to the user.
 
+    :  <dfn>paymentEntitiesLogos</dfn> member
+    :: The logos, if any, that were displayed to the user representing entities that are facilitating the payment that this SPC call is for.
+
```
ianbjacobs, line 1310:
```suggestion
    :  <dfn>paymentEntitiesLogos</dfn> member
    :: The logos, if any, that were displayed to the user in the transaction dialog. These logos are intended to represent the entities facilitating this transaction.
    
```

spec.bs
```diff
@@ -1277,6 +1364,41 @@ contains the following members:
...
+<xmp class="idl">
+    dictionary PaymentEntityLogo {
+        required USVString url;
+        required USVString label;
+    };
+</xmp>
+
+The {{PaymentEntityLogo}} dictionary contains information describing a
+logo for a payment entity that is facilitating the payment that the current
+Secure Payment Confirmation call is for. It contains the following members:
```
ianbjacobs, line 1378:
```suggestion
logo for a payment entity that is facilitating the current transaction. It contains the following members:
```

spec.bs

spec.bs

spec.bs

ianbjacobs (approved):

PR #296, opened 2025-05-20
[Spec] Add browser bound keys privacy and security considerations.

pejic:
This change adds privacy and security considerations related to browser bound keys. See also #286 for the implementation.

***
<a href="https://pr-preview.s3.amazonaws.com/pejic/secure-payment-confirmation/pull/296.html" title="Last updated on May 22, 2025, 8:30 PM UTC (2ad6dab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/296/72eb74e...pejic:2ad6dab.html" title="Last updated on May 22, 2025, 8:30 PM UTC (2ad6dab)">Diff</a>

PR #297, opened 2025-05-20
Add the TAG security and privacy self-review questionnaire for BBKs.

pejic:
Based the w3c [Self-Review Questionnaire: Security and Privacy](https://www.w3.org/TR/security-privacy-questionnaire/)

### DIFF
--- a/security-privacy-questionnaire-bbk.md
+++ b/security-privacy-questionnaire-bbk.md
@@ -1,0 +1,123 @@
+
+# Browser bound key (BBK) for Secure Payment Confirmation (SPC): Security and Privacy Questionnaire
+
+This questionnaire covers the BBK addition to SPC. See also the [Secure Payment
+Confirmation Questionnaire](security-privacy-questionnaire.md).
+
+2.1. What information does this feature expose, and for what purposes?
+
+* BBK public key: The public key of the BBK key pair that is randomly generated
+  and bound to a specific passkey, a specific browser instance and a specific device.
+* BBK signature: The signature using the BBK private key (not exposed).
+
+When a user agent has no suitable method for generating and storing the private
+key, no BBK is exposed.
+
+Browser bound keys enhance SPC for regulatory environments that require a
+possession factor as part of multifactor authentication. The consensus of the
+payments ecosystem is that synched passkeys do not meet requirements for
+possession factors, and so after discussion between the Web Authentication
+Working Group and the Web Payments Working Group, we are adding this feature to
+SPC instead of adding a general-purpose possession factor (via extension) to Web
+Authentication.
+
+2.2. Do features in your specification expose the minimum amount of information necessary to implement the intended functionality?
+
+Yes. Only the public key and signature are exposed.
+
+2.3. Do the features in your specification expose personal information, personally-identifiable information (PII), or information derived from either?
+
+No. The BBK public-private key pair is generated randomly.
+
+2.4. How do the features in your specification deal with sensitive information?
+
+N/A
+
+2.5. Does data exposed by your specification carry related but distinct information that may not be obvious to users?
+
+N/A
+
+2.6. Do the features in your specification introduce state that persists across browsing sessions?
+
+Yes. The BBK public-private key pair persists across sessions.
+
+2.7. Do the features in your specification expose information about the underlying platform to origins?
+
+Because BBKs may be implemented in a variety of ways, in general it may be
+difficult to deduce information about the underlying platform from the presence
+of a BBK public key and signature. However, in practice, if user agents only
+implement BBKs via secure elements, the availability of a secure element may be
+inferred by the presence of the BBK public key and signature.
+
+2.8. Does this specification allow an origin to send data to the underlying platform?
+
+No.
+
+2.9. Do features in this specification enable access to device sensors?
+
+No.
+
+2.10. Do features in this specification enable new script execution/loading mechanisms?
+
+No.
+
+2.11. Do features in this specification allow an origin to access other devices?
+
+No.
+
+2.12. Do features in this specification allow an origin some measure of control over a user agent’s native UI?
+
+No.
+
+2.13. What temporary identifiers do the features in this specification create or expose to the web?
+
+No temporary identifiers.
+
+However, the BBK public key is returned across successful Secure Payment
+Confirmation invocations (and therefore across sessions) that return the passkey
+credential ID that was passed in. In other words, the passkey credential ID must
+provided and user must accept the transaction before the BBK public key is
+exposed.
+
+2.14. How does this specification distinguish between behavior in first-party and third-party contexts?
+
+There is no difference. BBKs are returned in first-party and third-party Secure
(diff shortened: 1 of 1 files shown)


security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
+
+# Browser bound key (BBK) for Secure Payment Confirmation (SPC): Security and Privacy Questionnaire
+
+This questionnaire covers the BBK addition to SPC. See also the [Secure Payment
+Confirmation Questionnaire](security-privacy-questionnaire.md).
+
+2.1. What information does this feature expose, and for what purposes?
+
+* BBK public key: The public key of the BBK key pair that is randomly generated
+  per passkey and per browser instance.
```
ianbjacobs, line 10:
```suggestion
* BBK public key: The public key of the BBK key pair that is randomly generated
  and bound to a specific passkey, a specific browser instance, and a specific device.
```

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+N/A
+
+2.6. Do the features in your specification introduce state that persists across browsing sessions?
+
+Yes. The BBK public-private key pair persists across sessions.
+
+2.7. Do the features in your specification expose information about the underlying platform to origins?
+
+The availability of a secure element may be inferred by the presence/non
+presence of the BBK public key and signature.
```
ianbjacobs, line 39:
```suggestion
Because BBKs may be implemented in a variety of ways, in general it may be difficult to infer information about the underlying platform from the presence of a BBK public key and signature. However, in practice, if user agents only implement BBKs via secure elements, the availability of a secure element may be inferred by the presence of the BBK public key and signature. 
```

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+
+No.
+
+2.12. Do features in this specification allow an origin some measure of control over a user agent’s native UI?
+
+No.
+
+2.13. What temporary identifiers do the features in this specification create or expose to the web?
+
+No temporary identifiers.
```
ianbjacobs, line 63: Could one argue that the BBK public key is a temporary identifier? I may not understand exactly what "temporary identifier" refers to.

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+
+2.14. How does this specification distinguish between behavior in first-party and third-party contexts?
+
+A BBK is associated to a passkey (and browser), so in order for a merchant to
```
ianbjacobs, line 67:
```suggestion
A BBK is associated with a device/passkey/user agent instance, so in order for a merchant to
```

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+retrieve a BBK public key they must already have access to the passkey
+credential id which is already the case for Secure Payment Confirmation.
+
+For example: A merchant requests the user agent to show the Secure Payment
+Confirmation dialog by providing the passkey's credential id that the merchant
+received from the bank, issuer or payment system.
+
```
ianbjacobs, line 74: I'm not sure that this answer is responsive to the question. I think the answer is that for the BBK functionality, there is no difference in behavior between 1p and 3p contexts.

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+2.15. How do the features in this specification work in the context of a browser’s Private Browsing or Incognito mode?
+
+In private browsing mode
+* When a BBK for a given passkey exists, that BBK provides its public key and
```
ianbjacobs, line 78:
```suggestion
* If the user can access a passkey and there is an associated BBK, that BBK provides its public key and
```

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+  signature.
+* When a BBK for a given passkey does not exist, no BBK is generated and
+  persisted to avoid leaving a record on disk. And, no public key nor signature
```
ianbjacobs, line 81:
```suggestion
  persisted to avoid leaving a record on disk. And, no public key or signature
```

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
...
+  are provided.
+
+2.16. Does this specification have both "Security Considerations" and "Privacy Considerations" sections?
+
+Yes.
```
ianbjacobs, line 86: Suggest adding: At this time there are no specific security considerations for BBKs.

security-privacy-questionnaire-bbk.md
pejic: I don't believe the BBK public key is temporary when issuers need the public key to verify the signature, rather than being an arbitrary and reset-able identifier. Nonetheless users can take some steps to reset the BBK (and its public key) - such as users deleting and reinstalling their browser.

The questionnaire gives examples like TLS Channel ID, Session Tickets, IPv6 addresses, and GamePad identifiers. In these examples the identifiers are arbitrary; whereas, the public key is entangled with the signature.

None-the-less, I added a description of BBK public keys being exposed.

security-privacy-questionnaire-bbk.md
```diff
@@ -0,0 +1,110 @@
+
+# Browser bound key (BBK) for Secure Payment Confirmation (SPC): Security and Privacy Questionnaire
+
+This questionnaire covers the BBK addition to SPC. See also the [Secure Payment
+Confirmation Questionnaire](security-privacy-questionnaire.md).
+
+2.1. What information does this feature expose, and for what purposes?
+
```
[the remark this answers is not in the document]
pejic: Added.

security-privacy-questionnaire-bbk.md
pejic: Changed.

security-privacy-questionnaire-bbk.md
pejic: Changed.

security-privacy-questionnaire-bbk.md
pejic: Yes, you are correct, it is SPC that has the first-party and third-party considerations; however, BBKs operate the same way.

Reworded to indicate BBKs work the same in first-party and third-party context. And, directing readers to the same question in the SPC questionnaire.

security-privacy-questionnaire-bbk.md
pejic: There is a separate pull, #296, adding BBK considerations to the existing security and privacy sections.

security-privacy-questionnaire-bbk.md
pejic: Changed.

pejic (commented):
Thanks for the feedback, Ian.

ianbjacobs (approved):

PR #327, opened 2026-05-05
Create allowlist for webauthn extensions accessed via SPC

### LINKED ISSUE #326: [Security] SPC should disallow extensions that expose private relying party sign in data
Hello SPC friends! Nina from WebAuthn here.

SPC currently allows WebAuthn extensions on assertions obtained on a third-party context. Unfortunately, some WebAuthn extensions are considered sensitive and may store secrets that should only be revealed to the relying party.

Imagine, for example, that a bank uses WebAuthn to sign their users in, and that the same credential doubles as a payment credential. Being able to have credentials perform this double duty is a design goal of SPC. Now the bank launches a feature where users can send messages to each other that are end-to-end encrypted, and the secret is derived from a [WebAuthn PRF](https://w3c.github.io/webauthn/#prf-extension). Or, perhaps the bank stores the secret using [large blob](https://w3c.github.io/webauthn/#sctn-large-blob-extension). Third parties can now access that secret simply by obtaining an SPC assertion with the corresponding extension.

My recommendation would be to maintain a list of allowed extensions that can be revealed in a third party context and block the rest. This change is not backwards compatible but should limit the blast radius (there's no valid reason I can think of for a payment assertion to request either large blob or prf).

stephenmcgruer:
Fixes #326

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/327.html" title="Last updated on May 25, 2026, 7:41 PM UTC (7ad72e0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/327/6756707...7ad72e0.html" title="Last updated on May 25, 2026, 7:41 PM UTC (7ad72e0)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -41,6 +41,12 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
             text: internal slot
             text: internal method
 
+spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html
+    type: dfn
+        text: credProtect; url: sctn-credProtect-extension
+        text: minPinLength; url: sctn-minPinLength-extension
+        text: pinComplexityPolicy; url: sctn-pinComplexityPolicy-extension
+
 spec: i18n-glossary; urlPrefix: https://www.w3.org/TR/i18n-glossary/
     type: dfn
         text: locale; url: dfn-locale
@@ -80,6 +86,8 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: create credential async loop; url: CreateCred-async-loop
         text: credentialPublicKey; url: authdata-attestedcredentialdata-credentialpublickey
         text: base64url encoding; url: base64url-encoding
+        text: largeBlob; url: largeblob
+        text: credProp; url: credProp
 
 spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
@@ -852,7 +860,10 @@ Confirmation authentication (that is, one where the caller's [=origin=] is not
 the same as the [=origin=] of the [=Relying Party=] for the requested
 credentials) are:
 
-*   TODO
+* [=credProps=]
+* [=credProtect=]
+* [=minPinLength=]
+* [=pinComplexityPolicy=]
 
 ## Checking if Secure Payment Confirmation is available ## {#sctn-secure-payment-confirmation-available-api}
 
@@ -2366,11 +2377,10 @@ extensions, they might be able to access sensitive data belonging to the
 [=Relying Party=] that the [=Relying Party=] did not intend to share with the
 caller.
 
-For example, the `largeBlob` extension [[webauthn-3#sctn-large-blob-extension]]
-allows for storing and retrieving large amounts of data. If an attacker could
-initiate an SPC ceremony with the `largeBlob` extension, they might be able to
-read data that the [=Relying Party=] had stored there, assuming the user
-authorizes the ceremony.
+For example, the [=largeBlob=] extension allows for storing and retrieving
+large amounts of data. If an attacker could initiate an SPC ceremony with the
+`largeBlob` extension, they might be able to read data that the [=Relying
+Party=] had stored there, assuming the user authorizes the ceremony.
 
 To mitigate this risk, Secure Payment Confirmation restricts the use of
 WebAuthn extensions in cross-origin authentication ceremonies to a small
--- a/spec.bs
+++ b/spec.bs
@@ -44,8 +44,8 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
 spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html
     type: dfn
         text: credProtect; url: sctn-credProtect-extension
-        text: minPinLength; url: sctn-minPinLength-extension
-        text: pinComplexityPolicy; url: sctn-pinComplexityPolicy-extension
+        text: minPinLength; url: sctn-minpinlength-extension
+        text: pinComplexityPolicy; url: sctn-pincomplexitypolicy-extension
 
 spec: i18n-glossary; urlPrefix: https://www.w3.org/TR/i18n-glossary/
     type: dfn
@@ -87,7 +87,7 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: credentialPublicKey; url: authdata-attestedcredentialdata-credentialpublickey
         text: base64url encoding; url: base64url-encoding
         text: largeBlob; url: largeblob
-        text: credProp; url: credProp
+        text: credProps; url: credprops
 
 spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
@@ -856,9 +856,7 @@ The [=payment method additional data type=] for this payment method is
 ## Allowed WebAuthn Extensions ## {#sctn-allowed-webauthn-extensions}
 
 The <dfn>allowed WebAuthn extensions</dfn> for a third-party Secure Payment
-Confirmation authentication (that is, one where the caller's [=origin=] is not
-the same as the [=origin=] of the [=Relying Party=] for the requested
-credentials) are:
+Confirmation authentication are:
 
 * [=credProps=]
 * [=credProtect=]
--- a/spec.bs
+++ b/spec.bs
@@ -816,7 +816,9 @@ members:
     :: Any [=WebAuthn extensions=] that should be used for the passed
         credential(s). The caller does not need to specify the
         [[#sctn-payment-extension-registration| payment extension]]; it is added
-        automatically.
+        automatically. For a third-party SPC call, only
+        [[#sctn-allowed-webauthn-extensions|allowed WebAuthn extensions]] may
+        be specified.
 
     : {{SecurePaymentConfirmationRequest/browserBoundPubKeyCredParams}}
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -843,6 +845,15 @@ members:
...
 {{SecurePaymentConfirmationRequest}}.
 
+## Allowed WebAuthn Extensions ## {#sctn-allowed-webauthn-extensions}
+
+The <dfn>allowed WebAuthn extensions</dfn> for a third-party Secure Payment
+Confirmation authentication (that is, one where the caller's [=origin=] is not
+the same as the [=origin=] of the [=Relying Party=] for the requested
+credentials) are:
+
+*   TODO
```
stephenmcgruer: @nsatragno - any thoughts on extensions to allow to start with? We can keep the list short and add them as they are required by partners, but I think we should also start with some 'obvious' ones... whichever those are! :D

nsatragno: Generally useful:
- [credProps](https://w3c.github.io/webauthn/#sctn-authenticator-credential-properties-extension)

If interested in supporting security keys:
- [credProtect](https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-credProtect-extension)
- [minPinLength](https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-minpinlength-extension)
- [pinComplexityPolicy](https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-pincomplexitypolicy-extension)

These should all be safe to expose.

Avoid `prf`, `largeBlob` & friends, and `credBlob`.

spec.bs
```diff
@@ -1082,6 +1093,18 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
         1. If |logo|["{{PaymentEntityLogo/label}}"] is empty, throw a
             {{TypeError}}.
 
+1. If |data|["{{SecurePaymentConfirmationRequest/extensions}}"] [=map/exists=]
+    and |data|["{{SecurePaymentConfirmationRequest/rpId}}"] is not the
+    [=origin=] of the [=relevant settings object=] of |request|:
```
nsatragno, line 1107: This seems more restrictive than what you want.

`subdomain.example.com` can claim `example.com` on a regular WebAuthn request. Shouldn't it be allowed to pass all the same extensions as in a normal WebAuthn request?

What about [related origins](https://w3c.github.io/webauthn/#sctn-related-origins)?

spec.bs
stephenmcgruer: Thank you, done!

spec.bs
stephenmcgruer: Interesting, I don't think I was aware that WebAuthn allowed this. Isn't that risky for hosting domains? I guess they just have to never have passkeys on `origin.com` and have to use a reserved `login.origin.com` sub-domain for their passkeys?

This is a wider spec bug then, see https://w3c.github.io/secure-payment-confirmation/#ref-for-steps-to-silently-determine-if-an-spc-credential-is-third-party-enabled . I will file an issue, thank you!

stephenmcgruer: https://github.com/w3c/secure-payment-confirmation/issues/328; if ok, I will consider that issue orthogonal to this PR as its an existing spec problem.

nsatragno: > Isn't that risky for hosting domains? 

We go up to eTLD+1, backed by the public suffix list. Relying on the PSL is not perfect, but seems good enough.

nsatragno (approved):
Approved, thank you!

spec.bs
```diff
@@ -2330,6 +2364,28 @@ an unauthorized payment using Secure Payment Confirmation credentials
         if an attacker attempts to sit between the user and a valid merchant
         site and forward the assertion.
 
+### Restricted WebAuthn Extensions ### {#sctn-security-restricted-webauthn-extensions}
+
+WebAuthn allows for [=authentication extensions=] to be used during an
+authentication ceremony. Some extensions can be used to store or retrieve data
+on the authenticator that is sensitive to the [=Relying Party=].
```
ianbjacobs:
```suggestion
(on the authenticator) that is sensitive to the [=Relying Party=].
```

spec.bs
```diff
@@ -2330,6 +2364,28 @@ an unauthorized payment using Secure Payment Confirmation credentials
...
+caller.
+
+For example, the [=largeBlob=] extension allows for storing and retrieving
+large amounts of data. If an attacker could initiate an SPC ceremony with the
+`largeBlob` extension, they might be able to read data that the [=Relying
+Party=] had stored there, assuming the user authorizes the ceremony.
+
+To mitigate this risk, Secure Payment Confirmation restricts the use of
+WebAuthn extensions in cross-origin authentication ceremonies to a small
+[[#sctn-allowed-webauthn-extensions|allowlist of safe extensions]].
```
ianbjacobs:
```suggestion
[[#sctn-allowed-webauthn-extensions|list of allowed extensions]].
```

ianbjacobs: I propose this editorial change because the concept of "safe" has not been used previously, and the section header of the anchor does not use the word "safe."

ianbjacobs (approved):

spec.bs

spec.bs

spec.bs
```diff
@@ -80,6 +86,8 @@ spec: web-authn; urlPrefix: https://w3c.github.io/webauthn/
         text: create credential async loop; url: CreateCred-async-loop
         text: credentialPublicKey; url: authdata-attestedcredentialdata-credentialpublickey
         text: base64url encoding; url: base64url-encoding
+        text: largeBlob; url: largeblob
+        text: credProp; url: credProp
```
pejic: The link for `[=credProp=]` links me to `https://w3c.github.io/webauthn/#credProp` - an anchor that does not seem to exist. Maybe it should be `sctn-authenticator-credential-properties-extension` https://w3c.github.io/webauthn/#sctn-authenticator-credential-properties-extension?

spec.bs
```diff
@@ -41,6 +41,12 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
             text: internal slot
             text: internal method
 
+spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html
+    type: dfn
+        text: credProtect; url: sctn-credProtect-extension
+        text: minPinLength; url: sctn-minPinLength-extension
+        text: pinComplexityPolicy; url: sctn-pinComplexityPolicy-extension
```
pejic: Lowercase for this one too: `sctn-pincomplexitypolicy-extension`
https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-pincomplexitypolicy-extension

spec.bs
```diff
@@ -41,6 +41,12 @@ spec: ECMAScript; urlPrefix: https://tc39.github.io/ecma262/#
             text: internal slot
             text: internal method
 
+spec: FIDO-CTAP; urlPrefix: https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html
+    type: dfn
+        text: credProtect; url: sctn-credProtect-extension
+        text: minPinLength; url: sctn-minPinLength-extension
```
pejic: Only lowercase version navigates to the section. `sctn-minpinlength-extension` https://fidoalliance.org/specs/fido-v2.3-ps-20260226/fido-client-to-authenticator-protocol-v2.3-ps-20260226.html#sctn-minpinlength-extension

spec.bs
```diff
@@ -843,6 +853,18 @@ members:
 The [=payment method additional data type=] for this payment method is
 {{SecurePaymentConfirmationRequest}}.
 
+## Allowed WebAuthn Extensions ## {#sctn-allowed-webauthn-extensions}
+
+The <dfn>allowed WebAuthn extensions</dfn> for a third-party Secure Payment
+Confirmation authentication (that is, one where the caller's [=origin=] is not
+the same as the [=origin=] of the [=Relying Party=] for the requested
+credentials) are:
```
pejic: Should the `(that is, one where the caller's [=origin=] is not the same as the [=origin=] of the [=Relying Party=] for the requested credentials)` text be updated to include the part about the relevant settings object of request as below?

Alternatively, avoid the the duplicate statement here by referring to the "4.9. Steps to validate payment method data" section instead?

spec.bs
stephenmcgruer: Whoops, fixed. Thank you for catching that; it was a missing 's' at the end.

spec.bs
stephenmcgruer: Huh, TIL that anchors are case sensitive. Fixed.

spec.bs
stephenmcgruer: Done. (Love that `credProtect` however has to be uppercase 🤣 )

spec.bs
stephenmcgruer: Actually, let me just remove this 'clarification'. If we want 'third-party SPC authentication' to be clearer in the future, we should create a Terminology entry for it. For now, I think its fine without clarification.

pejic: Yes, removing the clarification in favor of third-party SPC authentication works too.

pejic (approved):

stephenmcgruer:
Per https://github.com/w3c/secure-payment-confirmation/issues/326#issuecomment-4537275441, I realized that the list of extensions makes no sense, so I reverted this commit in https://github.com/w3c/secure-payment-confirmation/commit/31cdf436f43db0573c9466dd46e5d126be73484b

PR #330, opened 2026-05-28
Bbk req update

John-Earnshaw:
Updates BBK requirements by changing "browser instance" to "installed browser program", with an appropriate definition
Appends an FAQ section to clarify behaviour for some specific scenarios 
Addresses concerns in #321


bbk-requirements.md
```diff
@@ -17,15 +17,15 @@ Therefore, the Web Payments Working Group plans to add a "browser-bound key (BBK
 ## Assumptions
 
 * Device binding is the fundamental property of a BBK. Once a BBK has been associated with a device, it cannot be reused outside that device.
-* Although in theory a BBK could be shared among multiple browser instances on the same device, we assume here that a BBK will also be bound to a browser instance (as the name "browser bound key" indicates).
+* Although in theory a BBK could be shared among multiple [installed browser programs](#installed-browser-program) on the same device, we assume here that a BBK will also be bound to an [installed browser program](#installed-browser-program) (as the name "browser bound key" indicates).
```
stephenmcgruer: This assumption sticks out to me on a re-read, @ianbjacobs . Do you know what we're trying to say here by "Although in theory a BBK could be shared [..]"?

Is it not a **requirement** rather than an **assumption** that BBKs are bound to a single installed browser program?

stephenmcgruer (approved):
LGTM, with one comment that is not technically related to this change but just highlighted by it!

bbk-requirements.md
```diff
@@ -77,6 +77,18 @@ _This section is in development._
 
 * This proposal relies on underlying [FIDO security assumptions](https://fidoalliance.org/specs/common-specs/fido-security-ref-v2.1-ps-20220523.html#fido-security-assumptions) and related threat models.
 
+## FAQ
+* Can a BBK be reused across different accounts on the same installed browser program?
+  * Implementation specific.
```
ianbjacobs:
```suggestion
  * This is an implementation choice left to the installed browser program.
```

bbk-requirements.md
```diff
@@ -77,6 +77,18 @@ _This section is in development._
...
+* Can a BBK be reused across an installed browser program and a web view version of that browser?
```
ianbjacobs:
```suggestion
* Can a BBK be reused across an installed browser program and a WebView version of that browser?
```

bbk-requirements.md
```diff
@@ -77,6 +77,18 @@ _This section is in development._
...
+  * No; each webview is considered a distinct installed browser program bounded by the embedding application.
```
ianbjacobs:
```suggestion
  * No; each WebView is considered a distinct installed browser program bounded by the embedding application.
```

bbk-requirements.md
```diff
@@ -77,6 +77,18 @@ _This section is in development._
...
+* Can a BBK be reused across an installed browser program and other installed browser programs using the same underlying engine (eg. Chromium)
```
ianbjacobs:
```suggestion
* Can a BBK be reused across an installed browser program and other installed browser programs using the same underlying engine (e.g., Chromium)?
```

bbk-requirements.md
```diff
@@ -17,15 +17,15 @@ Therefore, the Web Payments Working Group plans to add a "browser-bound key (BBK
...
 
 ## Definitions
 
-* <dfn id="bbk-binding">BBK binding</dfn>: a unique triplet consisting of a specific passkey, a specific browser instance, and a specific device.
-
+* <dfn id="bbk-binding">BBK binding</dfn>: a unique triplet consisting of a specific passkey, a specific [installed browser program](#installed-browser-program), and a specific device.
+* <dfn id="installed-browser-program">Installed browser program</dfn>: a specific installation of a web browser on a device. Note that a browser embedded in another application (e.g. WebView)  is considered a part of the embedding application and thus a distinct installed browser program.
```
ianbjacobs:
```suggestion
* <dfn id="installed-browser-program">Installed browser program</dfn>: a specific installation of a web browser on a device. Note that a browser embedded in another application (e.g., WebView)  is considered a part of the embedding application and thus a distinct installed browser program.
```

bbk-requirements.md
ianbjacobs: I agree that with this recent round of clarifications, we have elevated this to a requirement. I think we can delete this bullet.

bbk-requirements.md
ianbjacobs: I think we should put the definition of Installed browser program before the definition of BBK binding (since the latter uses the former).

bbk-requirements.md
John-Earnshaw: Makes sense, I'll remove it

stephenmcgruer (approved):

ianbjacobs (approved):

PR #331, opened 2026-05-28
Update UX Guidelines for Payment entity logos with actual dimensions for desktop

darwinlyang:


stephenmcgruer:
Thanks for sending this! Can we add a note for Chrome versions <= M149, e.g. something like

**NOTE**: On Chrome for Desktop versions < 150, a bug meant that logos were being improperly resized. If you want to handle those Chrome versions, you must pass a logo of size XxY instead. You can use the [User-Agent Client Hints API](https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API) to detect the Chrome version for serving different icon sizes.

darwinlyang:
Updated with the note.

stephenmcgruer (approved):

PR #332, opened 2026-06-02
Disallow webauthn extensions in cross-origin authentication ceremonies

### LINKED ISSUE #326: [Security] SPC should disallow extensions that expose private relying party sign in data

stephenmcgruer:
This reverts commit 31cdf436f43db0573c9466dd46e5d126be73484b, re-landing a new version of 31cdf436f43db0573c9466dd46e5d126be73484b. Instead of an allowlist, this version of the change just bans all extensions in cross-origin ceremonies.

Fixes #326

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/332.html" title="Last updated on Jun 3, 2026, 2:27 PM UTC (0d75a25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/332/3c2d9f8...0d75a25.html" title="Last updated on Jun 3, 2026, 2:27 PM UTC (0d75a25)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -821,8 +821,7 @@ members:
 
         NOTE: A third-party SPC call will throw if this dictionary is
         non-empty, to avoid attacks on [=Relying Party=] extensions that can
-        store private data. See
-        [[#sctn-security-restricted-webauthn-extensions]].
+        store private data. See [[#sctn-security-webauthn-extensions]].
 
     : {{SecurePaymentConfirmationRequest/browserBoundPubKeyCredParams}}
     :: The list of allowed types of credential restricting the types of
@@ -2341,7 +2340,7 @@ an unauthorized payment using Secure Payment Confirmation credentials
         if an attacker attempts to sit between the user and a valid merchant
         site and forward the assertion.
 
-### Restricted WebAuthn Extensions ### {#sctn-security-restricted-webauthn-extensions}
+### WebAuthn Extensions Only Allowed for the Relying Party ### {#sctn-security-webauthn-extensions}
 
 WebAuthn allows for [=authentication extensions=] to be used during an
 authentication ceremony. Some extensions can be used to store or retrieve data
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
@nsatragno - PTAL!

spec.bs
```diff
@@ -2330,6 +2341,28 @@ an unauthorized payment using Secure Payment Confirmation credentials
         if an attacker attempts to sit between the user and a valid merchant
         site and forward the assertion.
 
+### Restricted WebAuthn Extensions ### {#sctn-security-restricted-webauthn-extensions}
```
ianbjacobs: I am supportive of this pull request but think that the title of this section could be refined to make even clearer the scope. For example: "WebAuthn Extensions Only Allowed for the Relying Party".

If this title (or similar) is used, then references to this section (and the anchor) should be updated.

ianbjacobs (approved):

pejic (approved):

spec.bs
stephenmcgruer: Done, thanks!

stephenmcgruer:
We just landed the change in Chromium to block extensions for third-party SPC calls - https://chromium-review.git.corp.google.com/c/chromium/src/+/7869360 . It will ship in Chrome M151 (releases end of July) unless we end up cherry-picking it back into an earlier release. I will update the WG if we cherry-pick it earlier.

PR #334, opened 2026-06-05
Bbk spec alignment

John-Earnshaw:
Here's a first pass at aligning the spec with the new BBK requirements document.
Based on changes made for Issue #321

***
<a href="https://pr-preview.s3.amazonaws.com/John-Earnshaw/secure-payment-confirmation/pull/334.html" title="Last updated on Jul 2, 2026, 10:54 AM UTC (b6459d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/334/bc0e942...John-Earnshaw:b6459d8.html" title="Last updated on Jul 2, 2026, 10:54 AM UTC (b6459d8)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1827,7 +1827,7 @@ Note: The requirements and design considerations for browser bound keys are desc
 The browser bound key store contains the following:
 
     : <dfn>browser_bound_map</dfn>
-    :: a map from [=Credential ID=] to a key pair identifier. Each key pair may be associated with at most one [=Credential ID=].
+    :: a map from [=Credential ID=] to a key pair identifier. Each key pair identifier may be associated with at most one [=Credential ID=].
 
     : <dfn>keypair_map</dfn>
     :: a map from a byte array, the key pair identifier, to a [=tuple=] of the
--- a/spec.bs
+++ b/spec.bs
@@ -632,7 +632,8 @@ try {
 : <dfn>Browser bound key</dfn>
 :: A public-private [=key pair=] that signs over the transaction details in
     addition to the WebAuthn credential and is tied to a single device by the
-    user agent.
+    user agent. The name was chosen as initial implementations of this feature
+    were created for browsers, but the concept is applicable to any user agent.
 
 : <dfn>Key Pair</dfn>
 :: A key pair is a [=struct=] consisting of <dfn for="key pair">public
@@ -1803,9 +1804,9 @@ fields:
 This is an internal component owned by the [=user agent=] that manages
 platform-dependant cryptographic key pairs including their association to [=SPC
 credentials=] (i.e. passkeys). Each browser bound key store is owned by a single
-installed browser program, binding it to both the device and the browser. Each
-installed browser program may hold multiple browser bound key stores, for
-example if the browser has multiple profiles.
+user agent, binding it to both the device and the browser. Each
+user agent may hold multiple browser bound key stores, for
+example if the user agent has multiple profiles.
 This section explains the <dfn>browser bound key
 store</dfn> and procedures for creating, binding, retrieving, and signing using
 the browser bound key pairs.
@@ -1823,6 +1824,9 @@ Note: The [=browser bound key store=] needs to generate key pairs, retrieve key
 Note: The requirements and design considerations for browser bound keys are described in the
     [BBK requirements document](https://github.com/w3c/secure-payment-confirmation/blob/main/bbk-requirements.md).
     The design of the browser bound key store and its procedures are based on those requirements and considerations.
+    Within the requirements document, the term "installed browser program" is used to refer to the user agent, as
+    the document was written with browser implementations in mind. However, the requirements and design considerations
+    are applicable to any user agent implementation.
 
 The browser bound key store contains the following:
 
--- a/spec.bs
+++ b/spec.bs
@@ -1804,7 +1804,7 @@ fields:
 This is an internal component owned by the [=user agent=] that manages
 platform-dependant cryptographic key pairs including their association to [=SPC
 credentials=] (i.e. passkeys). Each browser bound key store is owned by a single
-user agent, binding it to both the device and the browser. Each
+user agent which binds it to both the device and the browser. Each
 user agent may hold multiple browser bound key stores, for
 example if the user agent has multiple profiles.
 This section explains the <dfn>browser bound key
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
John - thanks for taking this on!

@darwinlyang @pejic - can you review this in my absence? Sorry to dump it on you!

spec.bs
```diff
@@ -1802,7 +1802,11 @@ fields:
 
 This is an internal component owned by the [=user agent=] that manages
 platform-dependant cryptographic key pairs including their association to [=SPC
-credentials=] (i.e. passkeys). This section explains the <dfn>browser bound key
+credentials=] (i.e. passkeys). Each browser bound key store is owned by a single
```
darwinlyang, line 1806: I think this is confusing because the previous sentence mentions that the BBK store is owned by the user agent. Although currently the user agents for BBK stores are only web browsers, I think we probably want to keep it as user agent to not limit the definition.

To add the change about multiple profiles, it would make more sense to phrase it like "For example, if the user agent is a web browser with profiles, a browser bound key store may be created for each profile."

spec.bs
```diff
@@ -1816,10 +1820,14 @@ Note: The [=browser bound key store=] needs to generate key pairs, retrieve key
...
 
+Note: The requirements and design considerations for browser bound keys are described in the
+    [BBK requirements document](https://github.com/w3c/secure-payment-confirmation/blob/main/bbk-requirements.md).
+    The design of the browser bound key store and its procedures are based on those requirements and considerations.
+
 The browser bound key store contains the following:
 
     : <dfn>browser_bound_map</dfn>
-    :: a map from [=Credential ID=] to a key pair identifier.
+    :: a map from [=Credential ID=] to a key pair identifier. Each key pair identifier may be associated with at most one [=Credential ID=].
```
pejic: This property would be a note in the spec. If I understand w3c spec language correctly, we lean toward the imperative procedures: I.e. the procedures must result in this property being true; otherwise, the procedures need updating.

spec.bs
pejic: The browser bound key store is owned by the user agent, and the definition of user agent includes profiles. So merely following these procedures implies a user on a different device would be considered a separate user agent with a separate browser bound key store. I am unsure whether a negative statement (i.e. "do not sync the private key to another device") could be coded into the SPC procedures. None the less, the clarification from our requirements doc could be part of the new note below.

If we have additional security considers then we may need a statement in the security considerations. For example, the device bound session credentials spec mentions user agent responsibilities for protecting private keys in the security considerations section.

For reference here is the definition of [user agent](https://infra.spec.whatwg.org/#user-agent):
> A user agent is any software entity that acts on behalf of a user, for example by retrieving and rendering web content and facilitating end user interaction with it. In specifications using the Infra Standard, the user agent is generally an instance of the client software that implements the specification. The client software itself is known as an implementation. A person can use many different [user agents](https://infra.spec.whatwg.org/#user-agent) in their day-to-day life, including by configuring an [implementation](https://infra.spec.whatwg.org/#implementation) to act as several [user agents](https://infra.spec.whatwg.org/#user-agent) at once, for example by using multiple profiles or the implementation’s private browsing mode.

The DBSC [security considerations](https://www.w3.org/TR/dbsc/#security-considerations):
> The goals of DBSC are to reduce session theft by offering an alternative to long-lived cookie bearer tokens and to allow session authentication to be based on private keys that can be better protected against exfiltration. This makes the internet safer for users in that it is less likely their identity is abused, as malware is forced to act locally and thus becomes easier to detect and mitigate.
> User agent implementations are responsible for selecting the best way to protect such private keys from exfiltration, taking into account types of malware threats that exist on different platforms, and the protection mechanisms that exist. That includes, but is not limited to, secure hardware such as TPMs or secure elements, OS provided APIs for secure key generation and management, or process isolation mechanisms that provide reasonable protection against locally running malware.
>
> As long as the session is valid and was registered to an uncompromised device, a host can thus know with cryptographic certainty that the session key is protected against malicious exfiltration to a different device.

John-Earnshaw: > I think this is confusing because the previous sentence mentions that the BBK store is owned by the user agent. Although currently the user agents for BBK stores are only web browsers, I think we probably want to keep it as user agent to not limit the definition.
> 
> To add the change about multiple profiles, it would make more sense to phrase it like "For example, if the user agent is a web browser with profiles, a browser bound key store may be created for each profile."

Good spot, those sentences don't match.
Given our recent updates to the BBK requirements, should we not use "installed browser program" in both sentences to be consistent with the requirements doc?

spec.bs
John-Earnshaw: Sounds good, I've moved it to a separate note

spec.bs
darwinlyang: I personally think we should keep "user agent" as it's used throughout the rest of the SPC spec and to keep things open to other types of user agents in the future but maybe it's looking too far into the future.

@ianbjacobs, what are your thoughts on this?

pejic: Would a definition of installed browser program, combining user agent and device, serve to bridge the gap between the requirements doc and the spec?

Then this section would always reference installed browser program instead of user agent directly.

ianbjacobs: I agree that we should use "user agent" generally. We could put an informative note in the BBK section that says (roughly): because the first implementation of this feature was in a browser, we called it BBK, but that is not meant to imply that it could not be used in other user agents that implement SPC. Furthermore, we talk about "installed browser programs" because we are focused on the current implementation in browsers..

John-Earnshaw: I've aligned the terms to all be "user agent". To clarify things I have added a note on the use of the word "browser" in the definition for BBK, and I have explained the gap in terminology in the note that links to the BBK requirements.

darwinlyang: Thanks John! For the line:

> Each browser bound key store is owned by a single user agent, binding it to both the device and the browser.

I think it would be better to be more clear that the binding is a result of the store being owned by the user agent:

```
Each browser bound key store is owned by a single user agent which binds it to both the device and the browser.
```

spec.bs
```diff
@@ -632,7 +632,8 @@ try {
 : <dfn>Browser bound key</dfn>
 :: A public-private [=key pair=] that signs over the transaction details in
     addition to the WebAuthn credential and is tied to a single device by the
-    user agent.
+    user agent. The name was chosen as initial implementations of this feature
```
ianbjacobs:
```suggestion
    user agent. <b>Note:</b> The name was chosen as initial implementations of this feature
```

ianbjacobs (approved):

darwinlyang (approved):

pejic (approved):
Thanks @John-Earnshaw

PR #338, opened 2026-07-28
Explainer for supporting connected roaming authenticators

stephenmcgruer:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/338.html" title="Last updated on Aug 6, 2026, 7:36 PM UTC (603bbd5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/338/2af79bc...603bbd5.html" title="Last updated on Aug 6, 2026, 7:36 PM UTC (603bbd5)">Diff</a>

PR #340, opened 2026-07-28
Explainer for supporting non-queryable authenticators

stephenmcgruer:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/pull/340.html" title="Last updated on Aug 6, 2026, 7:36 PM UTC (00e8aac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/340/2af79bc...00e8aac.html" title="Last updated on Aug 6, 2026, 7:36 PM UTC (00e8aac)">Diff</a>


stephenmcgruer:
@ianbjacobs - my suggestion is that we would merge this and https://github.com/w3c/secure-payment-confirmation/pull/338, with an understanding that that doesn't indicate any approval (the state is still "Proposal"), and then discuss in the WG. The reason to do it that way is that the merged view of the explainer is much nicer IMO than the "markdown preview" you get under "Files changed" which underlines most of the words -_-. The downside however is thats its much harder to comment on individual lines/sections after its landed than during a PR, so maybe its harder for folks to give feedback.

What do you think?

ianbjacobs:
This is a very helpful description and I would definitely merge it. It would be great to walk through it at an upcoming call.

Issue #343, opened 2026-08-11
[spec] SPC Locale Field to Return Error on Language Mismatch


darwinlyang:
# Proposal

SPC has a [locale field](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-locale) that can be optionally set by RPs to indicate language preferences. RPs can provide a list of [language tags](https://www.rfc-editor.org/info/rfc5646/) in descending order of priority in which the browser (user agent) will use to decide on the locale to be used for the SPC dialog.

In reality, browsers ignore this field as they tend to just use the browser language for browser features (e.g. web authn, settings) to provide a cohesive experience for users and because it is not feasible for the browser to support every language. In addition, the browser language is something that the user explicitly chooses to use and cannot be changed by sites whereas the locale provided from RPs can end up being a locale that the user does not understand.

For example, if a user is making a purchase on a foreign site in a foreign language, they would rather see the SPC dialog be in their native language than the foreign language. In this case, RPs would not want to override the browser language used in the SPC dialog with the foreign language of the site.

But this presents a problem as the SPC dialog may contain strings or images supplied by the RP:
* [PaymentCredentialInstrument](https://www.w3.org/TR/secure-payment-confirmation/#dictdef-paymentcredentialinstrument)
* [payeeName](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-payeename)
* [payeeOrigin](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-payeeorigin)
* [paymentEntitiesLogos](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-paymententitieslogos)

<img width="932" height="848" alt="RP Supplied Data Elements" src="https://github.com/user-attachments/assets/86310d67-f99d-4150-ac3b-a409318166cd" />

These RP supplied data elements may contain locale details that might not match with the locale of the SPC dialog and so the RP requires a way to supply the data elements in the correct locale.

To resolve for this, I propose that we update the spec to continue to use the `locale` field but to validate the languages provided and return an error if **none** of the languages provided match the browser language. This can be checked with the `PaymentRequest.canMakePayment` method. An example of how this would be used:

```javascript
const spcData = {
   locale: ['en-CA'],
   ...
};

const request = new PaymentRequest(
   [{supportedMethods: 'secure-payment-confirmation', data: spcData}],
[...]

PR #345, opened 2026-09-02
Update the Specification for Locale Validation

darwinlyang:
The goal is to update the specification so that the locale field in `SecurePaymentConfirmationRequest` will return an error when `PaymentRequest.canMakePayment()` is called if:
* the locale field is set and non-empty
* none of the locales provided match the locale used by the Secure Payment Confirmation dialog

See https://github.com/w3c/secure-payment-confirmation/issues/343 for additional discussions for this PR.

***
<a href="https://pr-preview.s3.amazonaws.com/darwinlyang/secure-payment-confirmation/pull/345.html" title="Last updated on Sep 11, 2026, 8:13 PM UTC (9c79925)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/secure-payment-confirmation/345/071f5b2...darwinlyang:9c79925.html" title="Last updated on Sep 11, 2026, 8:13 PM UTC (9c79925)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -2647,17 +2647,11 @@ already be exposed through other APIs (e.g. via [[HTML#language-preferences]]).
 
 # Accessibility Considerations # {#sctn-accessibility-considerations}
 
-User agents render the {{PaymentCredentialInstrument/icon}} and
-{{PaymentCredentialInstrument/displayName}} together. Relying parties ensure the
-accessibility of the icon presentation by providing sufficient information via
-the {{PaymentCredentialInstrument/displayName}} (e.g., if the icon represents a
-bank, by including the bank name in the
-{{PaymentCredentialInstrument/displayName}}).
+User agents render the {{PaymentCredentialInstrument/icon}} and {{PaymentCredentialInstrument/displayName}} together. Relying parties ensure the accessibility of the icon presentation by providing sufficient information via the {{PaymentCredentialInstrument/displayName}} (e.g., if the icon represents a bank, by including the bank name in the {{PaymentCredentialInstrument/displayName}}).
 
 User Agents implementing this specification should follow both
 [[webauthn-3#sctn-accessiblility-considerations|WebAuthn's Accessibility Considerations]]
-and
-[=payment request accessibility considerations|PaymentRequest's Accessibility Considerations=].
+and [=payment request accessibility considerations|PaymentRequest's Accessibility Considerations=].
 
 # Internationalization Considerations # {#sctn-i18n-considerations}
 
--- a/spec.bs
+++ b/spec.bs
@@ -1098,12 +1098,32 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
     the [=relevant settings object=] of |request|, throw a {{TypeError}}.
 
 1. If |data|["{{SecurePaymentConfirmationRequest/locale}}"] [=map/exists=] and
-    [=list/is not empty=]
+    [=list/is not empty=]:
+
+    1. Let |dialogLanguageTag| be the [=language tag=] used by the Secure
+        Payment Confirmation dialog.
+
+    1. Let |matched| be false.
 
     1. For each |tag| in |data|["{{SecurePaymentConfirmationRequest/locale}}"]:
 
         1. If |tag| is not a well-formed [[BCP47]] language tag, throw a
             {{TypeError}}.
+
+        1. If both |tag| and |dialogLanguageTag| contain more than one
+            [=language subtag=], and |tag| and |dialogLanguageTag| are an
+            [=ASCII case-insensitive=] match, set |matched| to true and break.
+
+        1. If at least one of |tag| or |dialogLanguageTag| consists only of a
+            primary [=language subtag=], and their primary [=language subtags=]
+            are an [=ASCII case-insensitive=] match, set |matched| to true and
+            break.
+
+    1. If |matched| is false, throw a "{{NotSupportedError}}" {{DOMException}}.
+
+    NOTE: Returning {{NotSupportedError}} when no match is found lets the
+          RP know that the dialog locale does not match any of the requested
+          locale preferences.
 
     NOTE: The {{SecurePaymentConfirmationRequest/locale}} is distinct from
           language or direction metadata associated with specific input
@@ -1123,36 +1143,6 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
   authentication-icon-data-url.https.html
   authentication-rejected.https.html
 </wpt>
-
-1. If |data|["{{SecurePaymentConfirmationRequest/locale}}"] [=map/exists=] and
-    [=list/is not empty=]:
-
-    1. Let |dialogLanguageTag| be the [=language tag=] used by the Secure
-        Payment Confirmation dialog.
-
-    1. Let |matched| be false.
-
-    1. For each |tag| in |data|["{{SecurePaymentConfirmationRequest/locale}}"]:
-
-        1. If both |tag| and |dialogLanguageTag| contain more than one
-            [=language subtag=], and |tag| and |dialogLanguageTag| are an
-            [=ASCII case-insensitive=] match, set |matched| to true and break.
-
-        1. If at least one of |tag| or |dialogLanguageTag| consists only of a
-            primary [=language subtag=], and their primary [=language subtags=]
-            are an [=ASCII case-insensitive=] match, set |matched| to true and
-            break.
-
-    1. If |matched| is false:
-
-        1. Reject the
-            {{PaymentRequest/canMakePayment|PaymentRequest.canMakePayment()}}
-            promise with a "{{NotSupportedError}}" {{DOMException}}.
-
-        1. Return false.
-
-        Note: This lets the RP know that the locale used by the dialog does not
-        match their requested locale preference.
 
 1. If |data|["{{SecurePaymentConfirmationRequest/payeeOrigin}}"] is present:
 
--- a/spec.bs
+++ b/spec.bs
@@ -97,7 +97,9 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
 <pre class="link-defaults">
 spec:fetch; type:dfn; for:/; text:request;
 spec:i18n-glossary; type:dfn; text:bidi isolation
-spec:i18n-glossary; type:dfn; text:language priority list
+spec:i18n-glossary; type:dfn; text:language subtag
+spec:i18n-glossary; type:dfn; text:language tag
+spec:infra; type:dfn; text:ASCII case-insensitive
 spec:infra; type:dfn; text:list
 spec:infra; type:dfn; text:user agent
 spec:url; type:dfn; text:valid domain;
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -534,7 +536,7 @@ const request = new PaymentRequest([{
       },
     ],
 
-    // Caller’s requested localized experience
+    // Caller’s preferred locale(s) for validation
     locale: ["en"],
```
stephenmcgruer, line 540: We should add an example (as a subsection?) of what it looks like when locale matching fails

spec.bs
```diff
@@ -1119,6 +1124,36 @@ input {{SecurePaymentConfirmationRequest}} |data|, are:
   authentication-rejected.https.html
 </wpt>
 
+1. If |data|["{{SecurePaymentConfirmationRequest/locale}}"] [=map/exists=] and
```
stephenmcgruer, line 1127: This does not work to make canMakePayment() fail. It is confusing from the spec wording, but if you look at where the `steps to check if a payment can be made` are used in payment request, they're used *in show()*, not canMakePayment.

And you cannot just say "Reject the canMakePayment() promise" below, that does not work :). 

Unfortunately, we're running into API conflict here. `canMakePayment()` is meant to be agnostic to the specific data of a given payment request, it's general to "does the user agent have a handler that can handler the payment method *identifier*", not "this specific payment request".

I wonder if what we should be doing here is (ab)using hasEnrolledInstrument instead? What do you think?

(The annoying thing there is that hasEnrolledInstrument()) has been removed from the PaymentRequest spec even though its still present in the two browsers that ship PaymentRequest 🤦 , so it's a bit of a pain to spec. But we can still do it, and shouldn't let the spec pain stop us from doing the right semantic thing. Not that hasEnrolledInstrument is great semantics either...)

spec.bs
```diff
@@ -2599,36 +2629,60 @@ should make this clear to the user, for example with some clarifying text:
...
+If the locale used by the Secure Payment Confirmation dialog is relatively rare,
+it would provide a way for attackers to uniquely identify users.
+
+Although this is an avenue for fingerprinting, note that this locale may
+already be exposed through other APIs (e.g. via [[HTML#language-preferences]]).
+
 # Accessibility Considerations # {#sctn-accessibility-considerations}
 
-User agents render the {{PaymentCredentialInstrument/icon}} and {{PaymentCredentialInstrument/displayName}} together. Relying parties ensure the accessibility of the icon presentation by providing sufficient information via the {{PaymentCredentialInstrument/displayName}} (e.g., if the icon represents a bank, by including the bank name in the {{PaymentCredentialInstrument/displayName}}).
+User agents render the {{PaymentCredentialInstrument/icon}} and
```
stephenmcgruer: There appear to be a bunch of unrelated changes in this diff; please revert them and only make changes related to locale thanks!

stephenmcgruer (changes requested):

spec.bs
darwinlyang: Whoops, done.

spec.bs
darwinlyang: > they're used in show(), not canMakePayment

Should we consider rejecting all calls with invalid locales (including show)? If so, should this still be here in addition to another section on how this rejects on all calls?

> I wonder if what we should be doing here is (ab)using hasEnrolledInstrument instead? What do you think?

I don't mind adding `hasEnrolledInstrument` back in but I wonder if it's worthwhile to implement a new method `validatePaymentRequest` which could include other checks that can't be done in the renderer (e.g. if the page origin can claim the relying party ID).

spec.bs
darwinlyang: Ack. Will add this example when we finalize on how where we want to validate.

spec.bs
stephenmcgruer: > Should we consider rejecting all calls with invalid locales (including show)? If so, should this still be here in addition to another section on how this rejects on all calls?

I'm undecided for now. It likely depends on what other API we use for communicating the info. Generally speaking if `canMakePayment` returns false, I think `show()` should definitely throw. However if `hasEnrolledInstrument` returns false, I'm less certain...

> I don't mind adding hasEnrolledInstrument back in

To be clear, we don't need to add it back into Payment Request just to get it referenced here. That can be a bigger background work (see https://github.com/w3c/payment-request/pull/931, from... 2020).

> I wonder if it's worthwhile to implement a new method validatePaymentRequest which could include other checks that can't be done in the renderer (e.g. if the page origin can claim the relying party ID).

It's (mostly) not correct to think about concepts like renderer and browser when writing web specs. While it is true that we do shape APIs to what makes sense for engines to implement, they still have to make *semantic* sense for the website.

So the question you would then need to pose/answer is "What would `validatePaymentRequest` being true/false mean semantically, versus the existing `canMakePayment` and `hasEnrolledInstrument`?"

-----

Maybe `canMakePayment` would just be ok; we can acknowledge that we're slightly abusing the semantics for SPC. It's not like we don't do that already! 🤣 

Sorry I know there's no answer here, this is definitely an area where we just need to discuss and come to an agreement on.

spec.bs
darwinlyang: Closing this as this is now obsolete since it's part of the payment method data validation.

spec.bs
darwinlyang: Discussed offline and moving this to the payment request construction (under "Steps to validate payment method data").

This validation is very similar to Step 13 of the data validation:
> If data["[extensions](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-extensions)"] [exists](https://infra.spec.whatwg.org/#map-exists) and [is not empty](https://infra.spec.whatwg.org/#map-is-empty), and data["[rpId](https://www.w3.org/TR/secure-payment-confirmation/#dom-securepaymentconfirmationrequest-rpid)"] is not the [origin](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-origin) of the [relevant settings object](https://html.spec.whatwg.org/multipage/webappapis.html#relevant-settings-object) of request, throw a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror).

spec.bs
```diff
@@ -1093,13 +1098,33 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
     the [=relevant settings object=] of |request|, throw a {{TypeError}}.
 
 1. If |data|["{{SecurePaymentConfirmationRequest/locale}}"] [=map/exists=] and
-    [=list/is not empty=]
+    [=list/is not empty=]:
+
+    1. Let |dialogLanguageTag| be the [=language tag=] used by the Secure
```
stephenmcgruer, line 1103: I'm not 100% sure this is the right definition, but I leave this one to our i18n folks to be sure. (Maybe it should be "the language tag that the user agent will present the Secure Payment Confirmation dialog in"? Maybe we should be allowing for a *list* of tags?)

spec.bs
```diff
@@ -1093,13 +1098,33 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
...
+        Payment Confirmation dialog.
+
+    1. Let |matched| be false.
 
     1. For each |tag| in |data|["{{SecurePaymentConfirmationRequest/locale}}"]:
 
         1. If |tag| is not a well-formed [[BCP47]] language tag, throw a
             {{TypeError}}.
 
+        1. If both |tag| and |dialogLanguageTag| contain more than one
```
stephenmcgruer, line 1113: So IIUC, this handles the "en-us" === "en-us" !== "en-ca", is that correct?

spec.bs
```diff
@@ -1093,13 +1098,33 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
...
+            [=language subtag=], and |tag| and |dialogLanguageTag| are an
+            [=ASCII case-insensitive=] match, set |matched| to true and break.
+
+        1. If at least one of |tag| or |dialogLanguageTag| consists only of a
+            primary [=language subtag=], and their primary [=language subtags=]
```
stephenmcgruer: Is 'primary' a defined i18n concept? If so, can it be linked to?

spec.bs
```diff
@@ -1093,13 +1098,33 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
...
     1. For each |tag| in |data|["{{SecurePaymentConfirmationRequest/locale}}"]:
 
         1. If |tag| is not a well-formed [[BCP47]] language tag, throw a
             {{TypeError}}.
 
+        1. If both |tag| and |dialogLanguageTag| contain more than one
+            [=language subtag=], and |tag| and |dialogLanguageTag| are an
+            [=ASCII case-insensitive=] match, set |matched| to true and break.
+
+        1. If at least one of |tag| or |dialogLanguageTag| consists only of a
```
stephenmcgruer, line 1117: IIUC, this then handles both:

"en" === "en"

and

"en" === "en-us" or "en-ca", etc

Is that correct?

spec.bs
```diff
@@ -1093,13 +1098,33 @@ input {{PaymentRequest}} |request| and {{SecurePaymentConfirmationRequest}} |dat
...
+            primary [=language subtag=], and their primary [=language subtags=]
+            are an [=ASCII case-insensitive=] match, set |matched| to true and
+            break.
+
+    1. If |matched| is false, throw a "{{NotSupportedError}}" {{DOMException}}.
```
stephenmcgruer, line 1122: Do we handle wildcards? See https://w3c.github.io/i18n-glossary/#dfn-language-range 

I guess not since we're only allowing language *tags* and I think by https://www.rfc-editor.org/info/rfc5646/ anything with a `*` isn't a language tag?

spec.bs
darwinlyang: Yes, that is correct.

spec.bs
darwinlyang: Yes, that is correct.

spec.bs
darwinlyang: The current proposal does not allow for wildcards as `en` could just be used instead of `en-*`.

spec.bs
darwinlyang: Yes, I updated this to link to [BCP47](https://www.rfc-editor.org/info/rfc5646/#section-2.2.1).

spec.bs
stephenmcgruer: That makes sense, albeit I suspect there may be good reasons that our i18n community can think of to allow them (e.g., I believe `en-*-us` is allowed as a wildcard setup)

stephenmcgruer (approved):
I think this is ok from my viewpoint, although I imagine our i18n friends will ask if we have considered applying the matching algorithms from https://datatracker.ietf.org/doc/html/rfc4647#section-3, and why we haven't chosen them if so. It may be worth expressing that reasoning as a NOTE in the spec.

darwinlyang:
> I think this is ok from my viewpoint, although I imagine our i18n friends will ask if we have considered applying the matching algorithms from https://datatracker.ietf.org/doc/html/rfc4647#section-3, and why we haven't chosen them if so. It may be worth expressing that reasoning as a NOTE in the spec.

Sounds good. I added a note on why standard language negotiation isn't used.

